PR #60, opened 2017-05-30
On Premise and config file changes

marksou:
Changes to the On-Premises tutorials and config files. Some typo corrections as well.


tyk-docs/content/get-started/with-tyk-on-premise/index.md
```diff
@@ -1,6 +1,6 @@
 ---
 date: 2017-03-15T15:01:42Z
-title: With Tyk On Premise
+title: With Tyk On-Premise
```
ConsM, line 3: changing the title here means that you would need to change the `parent` key for it's "children" so they will sit inside.

PR #97, opened 2017-07-07
Proposed typos and formatting fixes in API Definition doc

cvillerm:


mvdan:
This looks good, thanks! Although I want @marksou's input regarding the american vs british spellings.

cvillerm:
Note that I kept british spelling, except when this was referring (already elsewhere) to something like an american-spelled header name

cvillerm:
Actually, when I changed "authorization code" to "authorisation" code, this was to keep the documentation consistent as there were other references to "authorisation code" elsewhere in the doc.

marksou:
@mvdan @cvillerm All looks good to me

mvdan:
Thanks again @cvillerm!

PR #110, opened 2017-07-24
OIDC: Setting use_openid to true in example

marians:
Seems weird to me that to enable OIDC the `use_openid` attribute would have to be set to `false`. Hoping for confirmation here from the Tyk team.

### DIFF
--- a/tyk-docs/content/security/your-apis/openid-connect.md
+++ b/tyk-docs/content/security/your-apis/openid-connect.md
@@ -50,7 +50,7 @@ To set up an API Definition to use OIDC, add the following block to the definiti
     "openid_options": {
         "providers": [
             {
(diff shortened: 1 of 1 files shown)


mvdan:
This page seems to say `true`: https://tyk.io/docs/integrate/open-id-connect/

Change seems fine to me, will let @buger confirm and press the button.

ConsM:
@marians thanks, good spot.
@mvdan it's actually set to `false` ? The object in both pages looks exactly the same, I will fix it on the Integrate section.

mvdan:
@ConsM you're right - I was looking at the other occurence in that page, which says `use_openid: Set to true to enable the OpenID Connect check.`

PR #124, opened 2017-08-25
Change Publish to Portal Management

marksou:
Decided that "Portal Management" was a better title for this section.


mvdan (approved):
Assuming that all the weight changes are on purpose, LGTM.

PR #137, opened 2017-09-13
Swapping Steps 6&7 because vim doesn't exist yet.

asoorm:


ConsM:
Good spot Ahmet, thanks !

PR #138, opened 2017-09-14
Deploy to Heroku hidden page update

marksou:
Created hidden page for Heroku clients


joshblakeley (approved):
Looks fine from my end, cheers.

PR #139, opened 2017-09-15
Update Redis Cluster info

marksou:
Updated the Redis Cluster page after discussion with Martin/Leon in response to Freshdesk # 1357


buger:
👍

PR #155, opened 2017-10-01
Tyk Pump Log File Update

huseyinbabal:
After on premise installation, there are only `tyk-pump.stderr` and `tyk-pump.stdout` not `tyk-pump.log`. I think this should be updated on doc


ConsM:
that is indeed the case, thank you.

Issue #169, opened 2017-10-10
Add users - example does not match. Should be POST


letzya:
https://tyk.io/docs/tyk-dashboard-api/users/
![image](https://user-images.githubusercontent.com/3155222/31400379-41e3ef62-ade7-11e7-9647-7b8343f5c246.png)

marksou:
Had spotted that earlier today. Is fixed in next push

marksou:
This is now fixed in the Docs.

PR #170, opened 2017-10-11
Various Updates

marksou:
Updated the Create Users page to clarify password issue. Also, fixes issue #169 
Updated Expiry and Quota settings issue.
Changed Tyk Component diagrams


tyk-docs/content/control-limit-traffic/token-expiry.md
```diff
@@ -21,7 +21,7 @@ Tyk manages timestamps in the Unix timestamp format - this means that when a dat
 
 Key sessions are created and updated using the Tyk REST API, in order to set the expiry date for a key, update the `expires` value with the timestamp of when the key should expire.
 
-Leave this field empty for a token never to expire.
+> **Note:** `expires` can only be a positive number, or -1 (unlimited). Zero (0) or other negative numbers, will not work.
```
mvdan, line 24: I'd say "will not work" is incorrect, as it is closer to undefined behavior. I didn't know empty string was documented before, so we should keep it defined as unlimited. So I'd write something like:

> `expires` can only be a positive number, or -1 (unlimited). Empty string will be interpreted as -1.

tyk-docs/content/get-started/with-tyk-cloud/tutorials/create-security-policy.md
```diff
@@ -47,7 +47,10 @@ A rate limit is enforced on all keys, set the number of requests per second that
 ![Quota form][5]
 
 Usage quotas limit the number of total requests a user is allowed to have over a longer period of time. So while a rate limit is a rolling window, a quota is an absolute maximum that a user is allowed to have over a week, a day or a month.
-**NOTE:** The Usage Quota set by a policy will override a quota applied to an individual key.
+
+Usage quotas can only be a positive number, or -1 (unlimited). Zero (0) or other negative numbers, will not work.
```
mvdan, line 51: Same as above - I'd just delete the last sentence. Or if you want to be explicit, say "undefined behavior".

mvdan (commented):
the comments apply to all "will not work" changes.

tyk-docs/content/control-limit-traffic/token-expiry.md
marksou: Changed all as requested

PR #207, opened 2017-11-23
SSO permission defaults - Update examples

asoorm:


marksou:
@buger @asoorm I'll update the sample code here as well? https://tyk.io/docs/dashboard-admin-api/sso/

asoorm:
Good spot. And whilst you are there - we can remove the tabs from the sample request too?

PR #213, opened 2017-11-27
Adds a note about Heroku Teams, fixes blockquotes

excieve:
The Heroku button only works with plans that support free Heroku dynos.
Blockquotes inside lists must be indented to be a part of a list item, otherwise it breaks list numbering.


marksou (approved):
Looks good.

PR #241, opened 2017-12-12
Transformation Output Fixed

huseyinbabal:
Transformed output in the example is wrong, `value-1` and `value-2` should be switched


marksou:
Can we check this is correct?

joshblakeley (approved):
LGTM, the output now matches what the above golang template is doing.

PR #245, opened 2017-12-16
clarify meaning of `cors.options_passthrough`

asoorm:
It seems from support and pre-sales enquiries that the wording on `options_passthrough` is a little confusing.

This PR aims to clarify the meaning of `cors.options_passthrough` config option.


tyk-docs/content/tyk-rest-api/api-definition-object-details.md
```diff
@@ -624,7 +624,9 @@ The CORS middleware has many options, these include:
 
 * `CORS.max_age`: Maximum age of credentials.
 
-* `CORS.options_passthrough`: Whether OPTIONS requests should pass through to the rest of the middleware. **Update in v1.9:** If this option is selected, OPTIONS requests will be completely ignored by the Tyk middleware. This means that pre-flight requests generated by web-clients such as SwaggerUI or the Tyk Portal documentation system will be able to test the API using trial keys.
+* `CORS.options_passthrough`: OptionsPassthrough instructs preflight to let next potential middleware to process the options method. Turn this on if your upstream handles `OPTIONS`.
```
lonelycode: Why is it in camel case? The option is in snake case in the config so we should be consistent.

The option doesn’t instruct preflight - that’s a browser controlled mechanism. 

Maybe;

```
This option ensures that browser CORS preflight requests (using the OPTIONS method), bypass the Tyk middleware chain so as to not interfere with an upstream-managed CORS configuration.

If your service handles CORS natively, then enable this option.
```

buger: I think we can do it even better: 
```
`CORS.options_passthrough` allow CORS OPTIONS preflight request to be proxied directly to upstream, without authentification and rest of checks. This means that pre-flight requests generated by web-clients such as SwaggerUI or the Tyk Portal documentation system will be able to test the API using trial keys.
```

marksou (approved):
Looks Good

PR #247, opened 2017-12-20
Add Azure support page

marksou:


buger:
I would probably not mention on-premise, but for seo reasons, I will rename title to: `Install Tyk API Gateway on Microsoft Azure`

Instead of mentioning quick start I would probably add link to our Docker docs instead https://tyk.io/docs/get-started/with-tyk-on-premise/installation/docker/

Also, urls to docs contain "localhost"

buger:
Also, You can install hybrid on Azure as well

PR #264, opened 2018-01-17
cleanups for clarity

joshblakeley:
use proper json notation in dashboard configuration page and gateway and dashboard env variables pages


marksou (approved):
Looks good.

PR #265, opened 2018-01-18
updated pump config example and added some notes

joshblakeley:
Layout changes, updated config to show more of the available pump configs. Mostly taken from pump repo readme


joshblakeley:
Layout looks funny in diff but fine on local.

marksou (approved):
Looks ok.

PR #269, opened 2018-01-22
Update plugin data structure: add HTTP method field

matiasinsaurralde:
Modifying [this page](https://tyk.io/docs/customise-tyk/plugins/rich-plugins/rich-plugins-data-structures/) to include the [new field](https://github.com/TykTechnologies/tyk/pull/1386).


marksou (approved):
Looks good.

PR #285, opened 2018-01-29
Adds a note on init systems, enables gpgcheck on rpm repos

excieve:
This adds a short guide on Linux init systems for the on-prem install. Describes their management differences and which OSes ship which one. Brings back package gpgcheck in the rpm repo files to make sure package signatures are verified by default, removes unused source repos from the repo files.


marksou (approved):
LGTM

PR #314, opened 2018-02-14
Draft for Hybrid on AWS installation

excieve:
This will be linked directly from AWS Marketplace product page but also available from the docs menu on Hybrid. This is draft because there's no public listing yet. Link to the product page might be also added later.


excieve:
@marksou Let's collaborate if this needs improvements.

excieve:
@james-tyk FYI

marksou (approved):
Looks good. We can hide the page from the menu if you want before it's listed on AWS Marketplace?

excieve:
@marksou I'd rather this doesn't get indexed by search engines as a docs page before we launch. And there will probably be some more related docs added. So maybe let's leave it as a draft for now? AFAIK Hugo doesn't publish drafts by default.

marksou:
@excieve Yes, that's correct.

PR #318, opened 2018-02-15
Update pump docs to include mongo aggregate

joshblakeley:
Adding this to example pump conf for cases when user might copy it in some way.


marksou (approved):
LGTM

PR #344, opened 2018-03-05
Update mdcb-setup.md

billxinli:
Updated JSON to use proper syntax and changed SSL default to false to reflect the previous part of the doc


buger:
Thank you!

I would rather update to use SSL everywhere. 

@marksou can you make update?

billxinli:
@buger Yep, agreed, ideally would be nice to keep both consistent, and if SSL it would also be nice to see how to set the below options:

```
"certificate": { "cert_file": <path>, "key_file": <path> },
"min_version": 771
```

buger:
Make sense!

PR #346, opened 2018-03-05
Update what-is-a-security-policy.md

letzya:
Have changed from client to user since that's the term we used above:

> Rate limit for a user

Also changed the wording a tiny bit


marksou (approved):
LGTM

PR #348, opened 2018-03-07
Add note about same function being available on global basis

marksou:
Following feedback


matiasinsaurralde (approved):
LGTM, I've updated the Python note a bit (doesn't affect Lua or gRPC).

PR #360, opened 2018-03-20
Fix TIB env var refs

mkinsley:
The TIB env var references were incorrect.  Fixing per feedback from community forum:

https://community.tyk.io/t/tib-unsupported-protocol-scheme/2476/2


buger:
Thank you!

PR #379, opened 2018-03-28
most important bit at top of doc.

asoorm:
Spent considerable time trying to work out why my response transform wasn't working.
If we need to enable response middleware in this manner, then should be specified at the top of the document


buger:
@asoorm but it is done automatically when you use dashboard, or I'm missing smth?

asoorm:
If not using dashboard - eg Tyk CE, or via dashboard api?

marksou (approved):
Looks good.

PR #382, opened 2018-03-30
Update batch-requests.md

lamari:
to be able to use Batch Requests I used the url in the doc "/{api-id}/tyk/batch" but it didnt work for me,  what did work is this "/{listen_path}/tyk/batch/"


marksou:
@joshblakeley Can you check this for me please?

PR #384, opened 2018-04-05
Release 2.6

marksou:


tyk-docs/content/configure/tyk-gateway-configuration-options.md
```diff
@@ -50,6 +50,21 @@ Set this value to `true` to enable key hashing, this will start hashing all keys
 
 If this is set to `true`, the same value should be enabled in the Dashboard configuration so that the UI can react appropriately.
 
+#### New for Gateway v2.6.0
```
buger: Are you sure this is not a mistake?  Seems like not valid page.

tyk-docs/content/security/security.md
```diff
@@ -6,7 +6,61 @@ menu: "main"
...
+*   Listing tokens is not possible
+*   Tokens appear in Analytics in their hashed form
+
+> **Warning**: Switching from a hashed installation to non-hashed means all existing tokens cannot be used (they will not be correctly validated).
+
+### <a name="tls-and-ssl"></a>TLS and SSL
+
+Tyk supports TLS connections and Mutual TLS. All TLS connections also support HTTP/2. Tyk also supports Let's Encrypt. See [TLS and SSL](/docs/security/tls-and-ssl/) for more details.
+
+### <a name="whitelisting"></a>Whitelisting
```
buger, line 48: Probably used in the wrong context. Whitelisting was mentioned in contexts of IP address whitelisting. And this section can be called Whitelisting and Blacklisting IPs.

tyk-docs/content/security/security.md
```diff
@@ -6,7 +6,61 @@ menu: "main"
...
+A hashed installation imposes some constraints on how Tyk is used:
+
+*   Listing tokens is not possible
+*   Tokens appear in Analytics in their hashed form
+
+> **Warning**: Switching from a hashed installation to non-hashed means all existing tokens cannot be used (they will not be correctly validated).
+
+### <a name="tls-and-ssl"></a>TLS and SSL
+
+Tyk supports TLS connections and Mutual TLS. All TLS connections also support HTTP/2. Tyk also supports Let's Encrypt. See [TLS and SSL](/docs/security/tls-and-ssl/) for more details.
```
buger: I think we should move Mutual TLS into the separate section. Right now this mix of terms do not really live together. 

Probably better to rewrite into smth like:

All Tyk components can be configured to use SSL or TLS for inbound or outbound connections. See [TLS and SSL] /docs/security/tls-and-ssl/. 

Probably does not make sense mention Lets encrypt there too, move it to tls-and-ssl page instead.

buger (approved):

tyk-docs/content/configure/tyk-gateway-configuration-options.md
marksou: Hi @buger I'm not sure which page you are referring to. Remember I've created new pages for some features.

buger: Well this content about new oauth endpoints lives in `tyk-gateway-configuration-options.md` which looks like mistake.

PR #407, opened 2018-04-30
Fix ID extractor configuration sample

matiasinsaurralde:
This configuration was using the value extractor, should be the regex extractor.


marksou (approved):
Looks good

PR #422, opened 2018-05-21
Deploying Tyk on-premise to Heroku guide

excieve:


marksou (approved):
Have made some edits, so it should be good to go. I'll have another look after it's gone live.

PR #434, opened 2018-06-01
Refactoring, move toc on the right

ConsM:


marksou (approved):
Ok, let us see how it affects things

PR #450, opened 2018-06-08
troubleshooting: fatal error dashboard & portal domains same

asoorm:
@marksou I have guessed how to create & linked pages - so if you could particularly check that this works, that would be great.


marksou (approved):
Yes, your link is correct format :-)

PR #465, opened 2018-06-20
Extend JSVM request data structure details

matiasinsaurralde:
Updated this based on the fields we've added in the latest release, `Scheme` is still missing here.


marksou:
Hi @matiasinsaurralde do you want to add the `scheme` info, before merging?

matiasinsaurralde:
Added `Scheme` with a sample.

PR #466, opened 2018-06-21
change close connections from true to false

asoorm:
This recommendation allows TCP re-use. Performance boost is
approximately 5x based on local tests


dencoded:
👍

marksou:
Are we happy to merge this now?

buger:
yes

PR #476, opened 2018-07-03
update caching page to reflect current code

joshblakeley:
Removed `options` from per path section as not in code atm


marksou (approved):
LGTM

PR #487, opened 2018-07-10
Hide nav container instead of the anchor tags

ConsM:
We were hiding the actual links instead of the nav, resulting in an odd UI


marksou (approved):
Looks good.

PR #575, opened 2018-10-23
SSL using certificate storage

buger:
Add way to configure Gateway SSL using certificate storage


marksou (approved):
Made a couple of edits. Good to go

PR #577, opened 2018-10-25
Added `proxy_close_connections` to optimised settings

davegarvey:
Discovered this setting based on something Leon wrote:

> Inside tyk.conf try setting `close_connections` value to `false`. 
> Option itself toggle keep-alive connection support, which basically means if new connection will be established for every request or connections is going to be re-used.
> 
> Before 2.6 this option was controlling both User -> Tyk and Tyk -> Upstream keep-alive behavior.
> 
> Starting from 2.6, this option was split into 2 parts: `close_connections` now controls only Us behavior, and new `proxy_close_connections` controls Tyk -> Upstream channel.


marksou (approved):
LGTM

PR #586, opened 2018-11-02
Added a note about the use of post method

letzya:


marksou (approved):
Changed case of Post. Good to go.

PR #598, opened 2018-11-23
Update capping-analytics-data-storage.md

letzya:
https://github.com/TykTechnologies/tyk/blob/8ecd114a20914729eb2abbdc1c101eee926d7067/cli/lint/schema.go#L260

should be `enforce_org_quotas` and not `enforce_org_quota` without an 's'. Have verified there's no such variable in the code base


marksou (approved):
I've spotted the same error in a couple of other places. I'll change those as part of the Org Quotas page re-work.

marksou:
@letzya We have this entry in the Gateway Environment variables table `enforce_org_quotas    TYK_GW_ENFORCEORGQUOTAS` Does this need changing as well?

We also have this entry in `tyk.conf` `"enforce_org_quotas": false,`

PR #601, opened 2018-11-26
Add delete key by hash option to Dashboard config

marksou:


tyk-docs/content/configure/tyk-dashboard-configuration-options.md
```diff
@@ -136,6 +137,8 @@ Each node communicates with the Dashboard via a shared secret (this setting) and
 
 *   `hash_keys`: If your Tyk Gateway is using hashed tokens, set this value here to `true` so it matches, the Dashboard will now operate in a mode that is compatible with key hashing.
 
+*   `enable_delete_key_by_hash`: To delete a key by its hash, this option needs to be set to `true`.
+
```
letzya, line 141: what do you think about:
Set to true if you want to allow your dashboard users to be able to delete keys that were found by a hash lookup. Needless to say that this option is relevant in case `hash_keys` was set to `true`.

ConsM: @letzya this option refers to API calls only, doesn't have to do with the UI (we perform a different check there)

PR #602, opened 2018-11-26
Update log-data.md

letzya:
Edit the doc a bit to make it clearer.
Added title for debug and for the integration with 3rd tooling


letzya:
@marksou that's a doc for the gw only can we make it clearer? maybe restructure and have a page per component - i.e. 
- [ ] log data for the dash 
- [ ] log data for pump

PR #628, opened 2018-12-18
Put env variables in alphabetic order

marksou:


asoorm (approved):

asoorm (changes requested):
If we going in alphabetical order, need it to be by 2nd column.

PR #669, opened 2019-02-05
Add note re default HTTP and HTTPS ports

marksou:


joshblakeley (changes requested):
Can you make the port in the example 8080 then were all gravy

PR #679, opened 2019-02-19
Add http2 options and updated tyk conf

marksou:
Updated gRPC intro page and tyk.conf settings. Need any Environment Variables to add to the Gateway Env Variables page.

### DIFF
--- a/tyk-docs/content/configure/tyk-gateway-configuration-options.md
+++ b/tyk-docs/content/configure/tyk-gateway-configuration-options.md
@@ -28,13 +28,13 @@ If `--conf` is not used, the first of the following paths to exist is used:
 
 Setting this value will change the port that Tyk listens on, by default Tyk will try to listen on port 8080.
 
-### <a name="secret"></a> secret
-
-This should be changed as soon as Tyk is installed on the system. This value is used in every interaction with the Tyk REST API, it should be passed along as the `X-Tyk-Authorization` header in any requests made. Tyk assumes that you are sensible enough not to expose the management endpoints to the public and to keep this configuration value to yourself.
-
 ### <a name="node_secret"></a> node_secret
 
 The shared secret between the Gateway and the Dashboard to ensure that API Definition downloads, heartbeat and Policy loads are from a valid source.
+
+### <a name="secret"></a> secret
+
+This should be changed as soon as Tyk is installed on the system. This value is used in every interaction with the Tyk REST API, it should be passed along as the `X-Tyk-Authorization` header in any requests made. Tyk assumes that you are sensible enough not to expose the management endpoints to the public and to keep this configuration value to yourself.
 
 ### <a name="template_path"></a> template_path
 
@@ -235,6 +235,7 @@ Set these options to hard-code values into the way the HTTP server behaves.
 
 ```
 "http_server_options": {
+  "enable_http2": false
   "override_defaults": false,
   "use_ssl": false,
   "enable_websockets": false,
@@ -254,6 +255,11 @@ Set these options to hard-code values into the way the HTTP server behaves.
   "ssl_insecure_skip_verify": false
 },
 ```
+
+#### <a name="enable_http2"></a> enable_http2
+
+Set to true for HTTP 2 connections.
+
 #### <a name="http_server_options-use_ssl"></a> http_server_options.use_ssl
 
 Set to `true` to enable SSL connections.
@@ -457,18 +463,22 @@ For additional security it is possible to have Tyk put its REST API on a separat
 
 > **NOTE**: This has been deprecated. Enter a value for `control_api_hostname` instead.
 
+### <a name="hostname"></a> hostname
 
-### <a name="enable_jsvm"></a> enable_jsvm
-
-By default we have now disabled the JavaScript middleware system to ensure higher performance on nodes. If you are using the JSVM (custom middleware, or virtual endpoints), then enable this setting.
+The hostname to bind the node to. If set, all API traffic must go via this host name, otherwise it will raise a 404.
 
 ### <a name="enable_custom_domains"></a> enable_custom_domains
 
 Set this value to `true` to enable this node to bind APIs to custom domains set in the API definition.
 
-### <a name="hostname"></a> hostname
+### <a name="proxy_enable_http2"></a> proxy_enable_http2
 
-The hostname to bind the node to. If set, all API traffic must go via this host name, otherwise it will raise a 404.
+Set this to `true` for HTTP 2 upstream connections.
+
+
+### <a name="enable_jsvm"></a> enable_jsvm
+
+By default we have now disabled the JavaScript middleware system to ensure higher performance on nodes. If you are using the JSVM (custom middleware, or virtual endpoints), then enable this setting.
 
 ### <a name="disable_virtual_path_blobs"></a> disable_virtual_path_blobs
 
--- a/tyk-docs/content/configure/tyk-gateway-configuration-options.md
+++ b/tyk-docs/content/configure/tyk-gateway-configuration-options.md
@@ -235,7 +235,7 @@ Set these options to hard-code values into the way the HTTP server behaves.
 
 ```
 "http_server_options": {
-  "enable_http2": false
+  "enable_http2": true,
   "override_defaults": false,
   "use_ssl": false,
   "enable_websockets": false,
@@ -258,7 +258,7 @@ Set these options to hard-code values into the way the HTTP server behaves.
 
 #### <a name="enable_http2"></a> enable_http2
 
-Set to true for HTTP 2 connections.
+This defaults to true for HTTP/2 connections.
 
 #### <a name="http_server_options-use_ssl"></a> http_server_options.use_ssl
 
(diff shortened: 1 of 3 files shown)


tyk-docs/content/configure/tyk-gateway-configuration-options.md
```diff
@@ -235,6 +235,7 @@ Set these options to hard-code values into the way the HTTP server behaves.
 
 ```
 "http_server_options": {
+  "enable_http2": false
```
furkansenharputlu: This value should be `true` by default. Also, there is a missing `,` at the end.

tyk-docs/content/configure/tyk-gateway-configuration-options.md
```diff
@@ -457,18 +463,22 @@ For additional security it is possible to have Tyk put its REST API on a separat
...
 
 ### <a name="enable_custom_domains"></a> enable_custom_domains
 
 Set this value to `true` to enable this node to bind APIs to custom domains set in the API definition.
 
-### <a name="hostname"></a> hostname
+### <a name="proxy_enable_http2"></a> proxy_enable_http2
 
-The hostname to bind the node to. If set, all API traffic must go via this host name, otherwise it will raise a 404.
+Set this to `true` for HTTP 2 upstream connections.
```
furkansenharputlu: Same for this. Its default value should be `true`. Also, `HTTP/2`.

tyk-docs/content/configure/tyk-gateway-configuration-options.md
```diff
@@ -254,6 +255,11 @@ Set these options to hard-code values into the way the HTTP server behaves.
   "ssl_insecure_skip_verify": false
 },
 ```
+
+#### <a name="enable_http2"></a> enable_http2
+
+Set to true for HTTP 2 connections.
```
furkansenharputlu: The general usage is `HTTP/2`.

tyk-docs/content/customise-tyk/plugins/rich-plugins/grpc/grpc.md
```diff
@@ -7,11 +7,30 @@ menu:
...
 
+### How can I use gRPC with Tyk?
+
+Using Tyk with your gRPC client and server is very easy. Since gRPC uses HTTP/2, you need to enable it by setting `enable_http2=true` for `Downstream-Tyk` and `proxy_enable_http2=true` for `Tyk-Upstream` connections. You also need to set your `listen_path` in your API definitions. 
+
+### Mutual Authentication
+Tyk supports Mutual Authentication in gRPC. See https://tyk.io/docs/security/tls-and-ssl/mutual-tls/ to configure Mutual Authentication in Tyk. Now, everything will be same as default Mutual Authentication.
+
+### Basic Authentication
+Tyk supports Basic Authentication in gRPC. See https://tyk.io/docs/security/your-apis/basic-auth/ to configure Basic Authentication in Tyk.
```
furkansenharputlu: Maybe we can hide links like this:
See [Basic Authentication](https://tyk.io/docs/security/your-apis/basic-auth/) section to configure ...

marksou: Changed in new commit

tyk-docs/content/configure/tyk-gateway-configuration-options.md
marksou: Changed in new commit

tyk-docs/content/configure/tyk-gateway-configuration-options.md
marksou: Changed in a new commit.

tyk-docs/content/configure/tyk-gateway-configuration-options.md
marksou: Changed in a new commit.

furkansenharputlu (approved):

furkansenharputlu:
@buger I think new env vars will be:
`TYK_GW_HTTPSERVEROPTIONS_ENABLEHTTP2`
`TYK_GW_PROXYENABLEHTTP2` 
right?

tyk-docs/content/configure/tyk-gateway-configuration-options.md
```diff
@@ -457,18 +463,22 @@ For additional security it is possible to have Tyk put its REST API on a separat
...
-### <a name="hostname"></a> hostname
+### <a name="proxy_enable_http2"></a> proxy_enable_http2
 
-The hostname to bind the node to. If set, all API traffic must go via this host name, otherwise it will raise a 404.
+This defaults to `true` for HTTP/2 upstream connections.
+
+
+### <a name="enable_jsvm"></a> enable_jsvm
+
+By default we have now disabled the JavaScript middleware system to ensure higher performance on nodes. If you are using the JSVM (custom middleware, or virtual endpoints), then enable this setting.
```
asoorm, line 481: > By default we have now disabled the JavaScript middleware system to ensure higher performance on nodes.

Does this not belong inside the changelog?

Could be reworded to:

`enable_jsvm` is false by default. Enable if you are using JSVM custom middleware or virtual endpoints.

tyk-docs/content/configure/tyk-gateway-configuration-options.md
```diff
@@ -254,6 +255,11 @@ Set these options to hard-code values into the way the HTTP server behaves.
   "ssl_insecure_skip_verify": false
 },
 ```
+
+#### <a name="enable_http2"></a> enable_http2
+
+This defaults to true for HTTP/2 connections.
```
asoorm, line 261: I don't understand what this means. config option `enable_http2` implies that I have to explicitly enable http2.

Yet the description `This defaults to true for HTTP/2 connections.` implies that true is the default. If `enable_http2: true` is default, then maybe config option should be `disable_http2`?

tyk-docs/content/customise-tyk/plugins/rich-plugins/grpc/grpc.md
```diff
@@ -7,11 +7,30 @@ menu:
...
 ---
-### What is gRPC?
+## What is gRPC?
 From the [gRPC documentation][1]:
 
 > gRPC is a modern, open source remote procedure call (RPC) framework that can run anywhere. It enables client and server applications to communicate transparently, and makes it easier to build connected systems.
 
+### How can I use gRPC with Tyk?
+
+Using Tyk with your gRPC client and server is very easy. Since gRPC uses HTTP/2, you need to enable it by setting `enable_http2=true` for `Downstream-Tyk` and `proxy_enable_http2=true` for `Tyk-Upstream` connections. You also need to set your `listen_path` in your API definitions.
```
asoorm, line 17: Are we able to use http2/gRPC insecure skip verify or without https? Is it just http or can the gRPC upstream server be tcp/unix socket? What does the connection string look like?

asoorm: Also, terminology for `Downstream-Tyk` and `Tyk-Upstream` is a bit confusing.

tyk-docs/content/configure/tyk-gateway-configuration-options.md
buger: Our default configs have this option enabled. So all new installations have it, and old will have to explicitly enable it.

tyk-docs/content/customise-tyk/plugins/rich-plugins/grpc/grpc.md
buger: Only http2 protocol supported, and http2 works only with TLS. And our standard TLS options will work.

buger: Agree about confusion. Can be re-worded to smth like this:

Tyk supports gRPC pass hough proxying when HTTP/2 used as a transport (the most common way to deploy gRPC services). So the only requirement is enabling HTTP/2 support on the gateway side, for both incoming and upstream connections, by setting `http_server_options.enable_http2` and `proxy_enable_http2` to `true`.

[gRPC over HTTP2 specification](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md) define rules on how gRPC protocol maps to HTTP request. In context of the API Gateway, we are intersted in the following things:
- HTTP path follow the format `/{Service-Name}/{method name}`, for example: `/google.pubsub.v2.PublisherService/CreateTopic`. You can use this feature to apply standard ACL rules via Keys and Policies, or use url rewrite plugins in endpoint desiger.
- HTTP method always `POST`
- gRPC custom request metadata added as HTTP headers, where metadata key directly mapped to HTTP header with the same name.

buger:
Whole updates from this PR inside plugins/rich-plugins/grpc/grpc.md should be moved to a separate page, because gRPC proxying do not have any relation to our gRPC plugins. This is a totally separate feature.

buger:
Needs additional paragraph on that Tyk can be used as a facade level, e.g. url rewriter can be used. So different services urls can point to different grpc servers. 

Also add code example for go-grpc, for basic auth part, showing that it just sets metadata.

PR #684, opened 2019-02-26
Fix some quirks with installation instructions

asoorm:


tyk-docs/content/get-started/with-tyk-on-premise/installation/on-ubuntu/dashboard.md
```diff
@@ -11,10 +11,6 @@ weight: 1
 
 Tyk has its own APT repositories hosted by the kind folks at [packagecloud.io][1], which makes it easy, safe and secure to install a trusted distribution of the Tyk Gateway stack.
 
-This tutorial will run on an Amazon AWS *Ubuntu Server 14.04 LTS* instance. We will install Tyk Dashboard with all dependencies stored locally.
```
excieve, line 14: Why remove this paragraph?

asoorm: Because i'm not sure it is relevant to quickstart. It's an ubuntu installation, and they know that from the title of the page. Can put it back if you think is useful?

excieve: I think it's useful as it sets `Ubuntu 14.04` as a reference point. Fine to remove AWS bits as they're indeed irrelevant.

asoorm: Maybe I got over-excited with the deleting. But the installation instructions works with Ubuntu 16 & 18 also.

excieve: You could add that it should also work on more recent versions.

excieve: But then the same should be done to other similar articles.

asoorm: I've made a change to suggest that it should work with newer Ubuntu versions with few modificatoins, if-any.

excieve: Thanks @asoorm! Also please check out `content/get-started/with-tyk-on-premise/installation/on-ubuntu/gateway.md` and `content/get-started/with-tyk-on-premise/installation/on-ubuntu/analytics-pump.md` — they have similar paragraphs. Although I'm not sure gateway fully works on more recent Ubuntu (thinking of Python compat).

asoorm: Didn't we fix the script to check which version of python was on the machine and use that?

https://github.com/TykTechnologies/tyk-analytics/blob/master/install/bootstrap.sh#L4

`PYTHON_BIN="$(command -v python || command -v python3)"`

excieve: I mean compat for Python lib in the gateway (3.4 vs. 3.6, etc.), not bootstrap script.

excieve (approved):
LGTM for the dashboard part.

marksou (approved):

PR #695, opened 2019-03-11
Made docs clearer for K8s users

lonelycode:
1. Made the Helm chart the preferred option for K8s
2. More explicit about how it works
3. Added a bit more terminology in there for findability


marksou (approved):
Looks good, just made a couple of minor edits.

PR #621, opened 2018-12-13
Release 2.8 docs update

buger:
Release notes draft https://github.com/TykTechnologies/tyk/releases/tag/untagged-d76c36c66ae8c8f5ba65

- [ ] Copy release notes to docs @marksou 
- [x] Dynamic portal customization part can go straight to this page https://tyk.io/docs/tyk-developer-portal/customise/customise-with-templates/
- [ ] Extend policy documents with "Separate rate limits and quotas per API within same Policy", including Policy API https://github.com/TykTechnologies/tyk/issues/356 @lonelycode 
- [ ] Add new page for API debugger (or section inside existing page) with screenshots. https://github.com/TykTechnologies/tyk-analytics-ui/issues/894 @buger 
- [ ] Add looping paragraph to URL rewriter section. And later separate page about more advanced usage, like SOAP support. May require a screenshot. https://github.com/TykTechnologies/product/issues/86 @buger 
- [ ] (Needs Release Note) Extend docs for Pump about new "hybird" pump, and in Hybrid docs, put note about possible configuration. https://github.com/TykTechnologies/tyk-pump/issues/80 @dencoded 
- [ ] Add note about new `health_check_endpoint_name` Gateway config field https://github.com/TykTechnologies/tyk/issues/1981 @buger 
- [ ] Update examples of using bundler CLI tools from to use `tyk bundler` instead of `tyk-cli bundler` https://github.com/TykTechnologies/tyk/issues/1492 @matiasinsaurralde 
- [ ] (Needs Release Note) Multi-org/multi-team support for enterprise https://github.com/TykTechnologies/product/issues/96 @joshblakeley 
- [x] Allow whitelist of acceptable ciphers https://github.com/TykTechnologies/tyk-analytics/issues/846 @asoorm 
- [x] Retrieve API details via public ID https://github.com/TykTechnologies/tyk-analytics/issues/1083 @adelowo 
- [x] Audit log improvements https://github.com/TykTechnologies/tyk-analytics/issues/344 @dencoded 
- [x] Request throttling https://github.com/TykTechnologies/tyk/issues/2089 @furkansenharputlu 
- [ ] Allow updating key by its hash https://github.com/TykTechnologies/tyk/issues/2055 @buger 
- [ ] (Needs Release Note) Proxy insecure skip verify support on per-api level https://github.com/TykTechnologies/tyk/issues/1837 @asoorm 
[...]

PR #717, opened 2019-04-02
document API details retrieval via public ID

adelowo:
See #621

This adds documentation for TykTechnologies/tyk-analytics#1083


marksou (approved):
Looks good

PR #742, opened 2019-04-12
Added note re normalising header

### LINKED ISSUE #740: Specify normalise form of header value is used while setting context variables for triggers
When advance URL Re-write feature is used, context variables set for matched values have the form 

`tyk_context.trigger-{n}-{name}-{i}`

where,
n = trigger index
name = regexp matcher name
i = index of that match

While using header match type in the trigger, the name is normalized form of header name. This should be specified explicitly.

marksou:
Fixes #740

### DIFF
--- a/tyk-docs/content/transform-traffic/url-rewriting.md
+++ b/tyk-docs/content/transform-traffic/url-rewriting.md
@@ -86,7 +86,8 @@ There are plenty of cases when path based rewriting is not enough. To cover this
 
 To make it work you should set the **triggers** field, defining rules. If there is no trigger match, the rewrite will fallback to the parent `rewrite_to`, but if either of the other two are triggered, the rewrite target is changed.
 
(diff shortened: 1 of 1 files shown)


tyk-docs/content/transform-traffic/url-rewriting.md
```diff
@@ -86,7 +86,8 @@ There are plenty of cases when path based rewriting is not enough. To cover this
 
 To make it work you should set the **triggers** field, defining rules. If there is no trigger match, the rewrite will fallback to the parent `rewrite_to`, but if either of the other two are triggered, the rewrite target is changed.
 
-Additionally, each trigger also sets a context variable for each match it finds. These context vars can then be used in the rewrites. Trigger contexts take the format: `$tyk_context.trigger-{n}-{name}-{i}` where `n` is the trigger index in the array, `name` is the regexp matcher name and `i` is the index of that match (since querystrings and headers can be arrays of values).
+Additionally, each trigger also sets a context variable for each match it finds. These context vars can then be used in the rewrites. Trigger contexts take the format: `$tyk_context.trigger-{n}-{name}-{i}` where `n` is the trigger index in the array, `name` is the regexp matcher name and `i` is the index of that match (since query strings and headers can be arrays of values).
+
```
komalsukhani, line 90: Can we move that Note here? Because it's not clear there that we are talking about the name component of the context variable.

marksou: Is this a better place?

PR #770, opened 2019-05-01
Response body update

marksou:
First draft


tyk-docs/content/transform-traffic/request-body.md
```diff
@@ -201,7 +201,7 @@ This mechanism operates the same way as the header injection middleware.
 
 ## <a name="request-body-context-data"></a> Context Data
 
-As of version 2.2 Tyk also allows context variables to be injected into the body using the `._tyk_context.` namespace, unlike the context exposed to the URL rewriter and header injector, the body transform can fully iterate through list indices so, for example, calling `_tyk_context.path_parts[0]` in a template will expose the first entry in the `path_parts` list.
+As of version 2.2 Tyk also allows context variables to be injected into the body using the `._tyk_context.` namespace. Unlike the context exposed to the URL rewriter and header injector, the body transform can fully iterate through list indices so, for example, calling `_tyk_context.path_parts[0]` in a template will expose the first entry in the `path_parts` list.
```
letzya: Not sure [0] is the right syntax.

letzya (changes requested):
The context var in body trans is repeated about 3 times out of which two mention iterating through a list (each had different syntax which makes it even more confusing).
(Just search the word “unlike” and see for yourself)
From brief review of the changes I think this page needs a bit of love - consolidate all the related info in one place and not in 3 different places in the same page but might be wrong.

tyk-docs/content/transform-traffic/request-body.md
marksou: Do we just need to remove the brackets?

asoorm: @marksou please try it - I think you need to specify `index _tyk_context.path_parts 0`

marksou: Fixed

PR #777, opened 2019-05-07
Add no-auth info

marksou:


tyk-docs/content/configure/outbound-email-configuration.md
```diff
@@ -47,6 +47,14 @@ To get email set up for your installation, add the following to your `tyk_analyt
 },
 ```
 
+From Tyk Dashboard v1.8.0, if `SMTPUsername` or `SMTPPassword` is omitted, Tyk assumes that authentication is not required for your SMTP server. When starting up and initialising the email driver, the Dashboard should output a log message as follows:
```
asoorm: For consistency, I think we should have:

```
#### SMTP NoAuth

> Available from Tyk Dashboard version 1.8

If `SMTPUsername` or `SMTPPassword` is omitted, ...
```

asoorm (changes requested):

tyk-docs/content/configure/outbound-email-configuration.md
marksou: Done

asoorm (approved):

PR #805, opened 2019-06-06
Env var changes

marksou:


tyk-docs/content/configure/environment-variables.md
```diff
@@ -0,0 +1,58 @@
...
+    parent: "Configure"
+weight: 12 
+---
+
+For each JSON based configuration option in the [Gateway](/docs/configure/tyk-gateway-configuration-options/), [Dashboard](/docs/configure/tyk-dashboard-configuration-options/) and [Pump](/docs/configure/tyk-pump-configuration/) configuration files, you can map to an Environment Variable. The Environment Variable will then override the configuration file option.
+
+The format for creating Environment Variables is as follows:
+
+* All Environment Variables are written in uppercase
+* All underscores in the JSON are removed in the environment variable
```
asoorm: Suggestion to change to:

* All underscores in the JSON key are removed in the environment variable
* All full stops in the JSON key are replaced by an underscore

example:

```
{
  "foo_bar.baz"
}
```
becomes

```
FOOBAR_BAZ
```

tyk-docs/content/configure/environment-variables.md
```diff
@@ -0,0 +1,58 @@
...
+* All full stops in the JSON are replaced by an underscore
```
asoorm: The full stops thing is misleading / confusing.

e.g. 

```
{
  "foo": {
    "bar": {
      "baz": "value"
    }
  }
}
```

to target `baz`, you you would do so using dot notation. `foo.bar.baz`.

Then to convert to an env var, you will change to uppercase and replace the dots with underscores.

`FOO_BAR_BAZ`

davegarvey (approved):

PR #818, opened 2019-06-19
Update the private key definition in certs

marksou:


letzya:
@marksou not sure I understand the pr.

letzya:
@marksou ?

marksou:
If I remember, @buger asked me to look at the wording. which I did.

letzya (approved):
Seems very similar. Hope I didn't miss anything

letzya (changes requested):
@marksou on 2nd thought - can you please explain me the actual difference? I don't understand it. thanks

buger:
The change here still make sense from grammar point of view, but it is not correct in a sentence about fallback. Using "private key" is not correct, is should say that it just falls back to `secret` field.

PR #820, opened 2019-06-20
Pro demo windows

marksou:
Page for running Pro Demo on Windows, and resurrected old page.


spicydarragh (approved):
Looks good!

I can put a document for an install with the Linux subsystem together but the only change from the original docker install file would be to run the setup file in the wsl terminal or to enter bash mode in Powershell and run it there. 

Windows needs to have Hyper-v enabled in order for Docker to run on it. VirtualBox needs to have it disabled in order to run. This may catch some people.

PR #839, opened 2019-07-09
Update Keys and Policy tutorials

### LINKED ISSUE #773: In tutorial - move the key creation after policy creation and fix the key creation content
This [doc](https://tyk.io/docs/get-started/with-tyk-on-premise/tutorials/tyk-on-premises-pro/create-api-token/) should also demo how to use policy when we creating a key.
At the moment it only adds an api to its access list which is correct but a better would be to first create a policy and then create a key with this policy. 
Using policies for keys is a very powerful feature that some other api gateways don't have

image

marksou:
Fixes #773


tyk-docs/static/include/create-api-key-include.md
```diff
@@ -14,15 +14,23 @@ We have a video walkthrough for creating an API Key.
 
 ![Add key button location][2]
 
-### Step 3: Set Access Rights
+### Step 3: Apply Policies (optional)
+
+This option allows the Key you create to be applied to many APIs at once. Policy settings override the individual Key settings. If you select a policy, all the APIs controlled by the policy will be listed in the **Access Rights** section.
+
```
letzya, line 20: Missing: the quota and rate limit set on the policy will be used by the key

PR #859, opened 2019-07-29
add default archeotype

gernest:
This allows easy generation of content files

now you can do this if you are in `tyk-docs` directory

```
hugo new new-content/new-content.md
```
Will  generate

```
./content/new-content/new-content.md
```

containing

```
---
title: "New Content"
date: 2019-07-29T10:28:52+03:00
weight: 0
draft: true
---

**Insert Lead paragraph here.**
```

Writing date by hand is error prone and not easy.

### DIFF
--- a/tyk-docs/archetypes/default.md
+++ b/tyk-docs/archetypes/default.md
@@ -1,0 +1,8 @@
+---
+title: "{{ replace .Name "-" " " | title }}"
+date: {{ .Date }}
+weight: 0
+draft: true
+---
+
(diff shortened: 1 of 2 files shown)


buger:
Worth noticing that you still need to set the proper parent, and probably remove draft option (since we are using pull requests anyway)

gernest:
@buger  I removed other fields that can/should be set by hand. `date` on the other hand should be automated . You can close this if it isn't useful but I find it handy now when trying to write docs.

PR #866, opened 2019-08-02
remove tcp connection recycling from prod docs

asoorm:
TCP connection recycling option removed from the kernel & it is not a best-practice / recommended approach.

See https://vincent.bernat.ch/en/blog/2014-tcp-time-wait-state-linux for further detail


excieve (approved):
:+1:

marksou (approved):

PR #868, opened 2019-08-05
Add troubleshooting page re port 5000 errors

marksou:


komalsukhani (approved):
I think Note should include `Port 5000 is only required if you need to enable the Tyk Gateway log viewer.`

PR #871, opened 2019-08-09
lint: Fixing page names & nav order

asoorm:
- Fix page names - e.g. remove "What is" from nav titles
- Removed duplicate page headings from content body
- Logical order for concepts pages

Screenshot 2019-08-09 at 12 51 31


marksou (approved):
LGTM

PR #875, opened 2019-08-09
changing the wording to make file handles section easier to find

sedkis:
changing the wording to make file handles section easier to find.  
no 2nd reviewers requested besides @marksou


marksou (approved):
LGTM

PR #883, opened 2019-08-15
Add Note re super user

marksou:


tyk-docs/content/dashboard-admin-api/users.md
```diff
@@ -74,6 +75,8 @@ admin-auth: 12345
 }
 ```
 
+> **NOTE**: You can also create a user without an `org_id`. This will create a "Super User", that won't be visible in the Users section of the Dashboard, but who can access APIs, etc, for all organisations created within Tyk.
```
sedkis: "...that won't be visible in the Users section of the Dashboard...."

this isn't accurate.  superusers can see themselves and all other users in the Users section. regular users cannot see superusers.

Maybe instead we can word like so:
"This will create a "Super User" who has global access to all APIs, Policies, etc, for all organisations created within Tyk."

sedkis: what do you think @marksou ?

sedkis (changes requested):
This will create a "Super User", that won't be visible in the Users section of the Dashboard, but who can access APIs, etc, for all organisations created within Tyk.

tyk-docs/content/dashboard-admin-api/users.md
marksou: Sure, good spot

marksou: Made change as requested.

sedkis (approved):

PR #901, opened 2019-09-25
adding information about how tyk uses custom auth sessions with caching

sedkis:


sedkis:
@marksou  mind checking off and letting one of the tech doods check off?

marksou (approved):
Looks ok to me. We can't merge until after 2.9 goes live as that's when we release Go Plugins :-)

buger:
I would be careful with putting Go plugin code as auth example, or at least provide Python/JSVM alternative.

sedkis:
> I would be careful with putting Go plugin code as auth example, or at least provide Python/JSVM alternative.

@buger @marksou This is via gRpc server ;)

buger:
I see but the doc is general, about auth plugins. 
gRCP is advanced topics.

sedkis:
@buger I don't know Python and JS auth is down.

Further, I would say that this isn't supposed to be an example of what to do.  It simply demonstrates the point that if you're creating a session object, just set the `LastUpdated` field so Tyk can properly handle this.

It's not meant to be an all out example so the programming language doesn't matter.

sedkis:
@marksou are we good to merge this with the other 2.9 stuff? can you add this to your list of things to go out?

marksou:
@sedkis Will do

PR #903, opened 2019-09-26
systemctl instead of service which doesn't work here

letzya:


alephnull (approved):
LGTM. Ubuntu is fully on board with systemctl, having gone through, by turn, invoke-rc.d, service and upstart.

marksou:
Going to merge this after speaking to @excieve. He also recommended removing any references to Ubuntu 14.04, as systemctl doesn't work on it, and is no longer supported.

PR #934, opened 2019-11-01
2.9.2 - pagination changes

sedkis:
updates for https://github.com/TykTechnologies/tyk-analytics/pull/1605

should be merged when the changes above go live

### DIFF
--- a/tyk-docs/content/tyk-dashboard-api/pagination.md
+++ b/tyk-docs/content/tyk-dashboard-api/pagination.md
@@ -9,9 +9,8 @@ weight: 1
 
 Select Dashboard APIs are paginated.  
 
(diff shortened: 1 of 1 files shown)


marksou (approved):
Tweaked it a little bit.

sedkis:
@marksou I don't think your changes are correct.  Via pagination, typically you describe which page exactly you wish to return, not the number of pages.
ie p=1 will return items 1-10
p=2 will return items 11-20
and so on

devsergiy (approved):

sedkis:
@marksou ,  not a big deal but we weren't supposed to merge this until 2.9.2 went live.  Now what we have in the public docs is completely incorrect

PR #937, opened 2019-11-04
Add some links to the Gateway API concept page

jlucktay:
When going through one of my new starter Trello cards, I got a bit stuck on one about the Gateway API, and some extra info and links like this would've cut right through the issue.


marksou (approved):
Looks good

PR #942, opened 2019-11-05
Fix broken link to Tyk On-Prem page

jlucktay:
Found that https://tyk.io/product/tyk-on-premises-free-edition/ goes to a 404.
I hope [the link I replaced it with](https://tyk.io/get-started/) is OK?

Did a quick pass over the rest of the page as well.


marksou (approved):
LGTM

PR #956, opened 2019-11-26
Added PUT endpoint for OAuth clients

marksou:
I've tested the YAML file in Swagger Editor

### DIFF
--- a/tyk-docs/static/others/gateway-swagger.yaml
+++ b/tyk-docs/static/others/gateway-swagger.yaml
@@ -1,7 +1,7 @@
 openapi: 3.0.0
 info:
   title: Tyk Gateway API
-  version: 2.9.2
+  version: 2.8.0
   description: |-
     The Tyk Gateway REST API is the primary means for integrating your application with the Tyk API Gateway system. This API is very small, and has no granular permissions system. It is intended to be used purely for internal automation and integration.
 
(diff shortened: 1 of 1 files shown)


tyk-docs/static/others/gateway-swagger.yaml
```diff
@@ -950,11 +950,34 @@ paths:
               example:
                 code: MWY0ZDRkMzktOTYwNi00NDRiLTk2YmQtOWQxOGQ3Mjc5Yzdk
                 redirect_to: 'http://client-app.com/oauth-redirect/?code=MWY0ZDRkMzktOTYwNi00NDRiLTk2YmQtOWQxOGQ3Mjc5Yzdk'
+    '/tyk/oauth/clients/{apiID}':
+    put:
+      summary: Update OAuth metadata
+      description: Allows you to update the metadata for an OAuth client.
+      tags:
+        - OAuth
+      operationId: listOAuthClients
```
buger: the wrong operationID, should be a unique value, like `updateoAuthClient`

Also, it is not about updating just metadata, you can also update policy id.

buger: Additionally, can we move it to the same section where we create oauth client?

marksou: Done

PR #961, opened 2019-11-29
Tried to rephrase the explanation and added an example

letzya:


marksou (approved):
LGTM

buger:
Looks good to me

PR #969, opened 2019-12-06
Mention 3.0 Swagger support

marksou:


joshblakeley (approved):
👍 

Maybe should also mention specifically that it is currently only json for OAS3 with yaml on roadmap

PR #977, opened 2019-12-12
Add supported Mongo and Redis versions.

marksou:


marksou:
Note this is just repeating the info we have on the front page, that isn't getting indexed by the search for some reason.

PR #985, opened 2019-12-19
update docs to include hmac request signature verification

adelowo:
Docs for https://github.com/TykTechnologies/tyk/pull/2452


marksou (approved):

buger:
@komalsukhani pls take ownership of this PR. We need to provide docs not only how we verify signatures, but also how we can sign upstream requests.

PR #1001, opened 2020-01-13
Add gRPC proxy info.

marksou:


davegarvey (commented):
I think that the topic of gRPC proxying should be separate from gRPC as a middleware. It doesn't feel right from an information architecture perspective.

We have a page for TCP proxy, so gRPC should be handled in the same way.

buger:
It is right. gRPC proxying should be separate section.

marksou:
@davegarvey  I've now created a new page.

marksou:
Given that this is caught up with the restructure, I'll add the new file into the new structure and merge, taking note of the other amended files

PR #1021, opened 2020-01-27
Create Logzio detailed info

joshblakeley:
We a lack of useful info about logzio pump so have fixed that.


marksou (approved):
LGTM

joshblakeley:
Cheers @marksou please merge

tyk-docs/content/tyk-configuration-reference/tyk-pump-configuration.md
```diff
@@ -338,6 +338,32 @@ Add the following section to expose the `/metrics` endpoint:
...
+```
+
+More advanced fields:
+
+`url` - If you do not want to use the default Logzio url i.e. when using a proxy. Default is `https://listener.logz.io:8071`
+`queue_dir` - The dir for the queue.
+`drain_duration` - Set drain duration (flush logs on disk). Default value is `3s`
+`disk_threshold` - Set disk queue threshold, once the threshold is crossed the sender will not enqueue the received logs. Default value is `98` (percentage of disk).
+`check_disk_space` - Set the sender to check if it crosses the maximum allowed disk usage. Default value is `true`.
+
```
letzya, line 389: Would be good to put it in another code example since it’s the fields are probably under logzio but one might think that some of them are under meta

letzya (changes requested):

tyk-docs/content/tyk-configuration-reference/tyk-pump-configuration.md
joshblakeley: They're all under meta actually so will add them for clarity

joshblakeley:
@letzya ptal

tyk-docs/content/tyk-configuration-reference/tyk-pump-configuration.md
```diff
@@ -358,6 +358,33 @@ Add the following section to expose the `/metrics` endpoint:
...
+```
+"logzio": {
+      "type": "logzio",
+      "meta": {
+        "token": "<YOUR-LOGZ.IO-TOKEN>"
+      }
+    }
+```
+
+More advanced fields:
```
letzya, line 379: can you put all these advanced fields in an example or declare they are under "meta". Don't think it's trivial to understand that

letzya (approved):

PR #1054, opened 2020-02-10
Updated inline with repo

marksou:


rewsmith (changes requested):
Prerequisites
Links to CE and Pro installs return 404's for me, but the installations are part of the guide, so aren't pre-reqs. A CE or Pro license is a prereq however.

Helm is also a prereq (text below copied from https://tyk.io/docs/getting-started/installation/with-tyk-on-premises/docker/with-kubernetes/ but excluded the tiller bits as tiller no longer exists in helm v3)

Helm 
If you are deploying dependency databases or Tyk components via Helm Charts, you will need to ensure that Helm is installed on your host / bastion machine

Installation: Community Edition
Mentions mongoDB, but AFAIK mongodb not needed/used in CE.  I would copy content from the GH helm chart repo (https://github.com/TykTechnologies/tyk-helm-chart#install-tyk-community-edition) instead.

Not really familiar enough with the rest of the content, but comparing against the GH helm chart repo it looks about right!

marksou:
Hi @rewsmith regarding MongoDB and CE. I spoke to @excieve about this, and although the CE doesn't use Mongo, the Pump will not work without it, so it was felt it should be mentioned. Will edit the Helm info. A licence is not required for the CE version.

PR #1111, opened 2020-03-10
Add internal plugin description

marksou:


tyk-docs/content/advanced-configuration/transform-traffic/endpoint-designer.md
```diff
@@ -85,6 +85,10 @@ By default the Ignore endpoint plugin is case-sensitive, so for example if `getu
 
 ![Blacklist options](/docs/img/dashboard/system-management/ignore_plugin_case.png)
 
+### Internal
```
letzya, line 88: it's not a plugin but a **config** and as such it shouldn't be in this page at all since it's in the "core setting" tab.

Also I would make the sentence a bit clearer to understand. Usually people are in a rush when breezing through docs and having 3 "to" and negative doesn't make it any easier "to set an endpoint to not be listened to by the "

letzya (changes requested):
per my comments

tyk-docs/content/advanced-configuration/transform-traffic/endpoint-designer.md
marksou: I'm confused now as it's listed as a plugin from the Endpoint Designer?

tyk-docs/content/advanced-configuration/transform-traffic/endpoint-designer.md
```diff
@@ -85,6 +85,10 @@ By default the Ignore endpoint plugin is case-sensitive, so for example if `getu
...
+
+This plugin allows you to set an endpoint to not be listened to by the Tyk Gateway, but can be called by other APIs by using the `tyk://` prefix.
```
buger: Update it to be `tyk://self/` prefix

tyk-docs/content/advanced-configuration/transform-traffic/endpoint-designer.md
buger: We have both API level config variable, and per endpoint. 
When you set it for the endpoint, it does not get exposed to the user, but you still can use it via looping.

tyk-docs/content/advanced-configuration/transform-traffic/endpoint-designer.md
marksou: @letzya are you happy to approve following @buger comments?

marksou:
Merging this as per @buger comments.

PR #1112, opened 2020-03-10
Add Downloads page

marksou:


tyk-docs/content/tyk-developer-portal/customise/downloads.md
```diff
@@ -0,0 +1,12 @@
+---
+title: "Downloads"
+date: 2020-03-10T18:09:52Z
+menu:
+  main:
+    parent: "Customise"
+weight: 7
+---
+
+If you want to have files available for download from your portal you can add them to your `/tyk-dashboard/portal/portal-assets` directory.
```
buger: The folder name should be `/opt/tyk-dashboard/...`

marksou: Have updated.

buger:
Do you think we need a new separate small page for this? 
Maybe it makes sense to embed it somewhere else, on the page about portal customization? (can be related to custom templates and custom CSS).

marksou:
I did wonder about that. I'll move it.

PR #1115, opened 2020-03-12
note about SSL

sedkis:


tyk-docs/content/basic-config-and-security/security/tls-and-ssl/tls-and-ssl.md
```diff
@@ -28,6 +28,8 @@ Tyk supports TLS connections, and as of version 2.0 all TLS connections will als
 
 You can enter multiple certificates, that link to multiple domain names, this enables you to have multiple SSL certs for your gateways or dashboard domains if they are providing access to different domains via the same IP.
 
+Finally, set the [host_config.generate_secure_paths](/docs/tyk-configuration-reference/tyk-dashboard-configuration-options/#a-namehost_configgenerate_secure_pathsahost_configgenerate_secure_paths) flag to `true` in your `tyk_analytics.conf`
```
letzya: The section above is for both tyk.conf and dashboard conf while this message is only for the dashboard so I would set the right context before "Finally..."

tyk-docs/content/basic-config-and-security/security/tls-and-ssl/tls-and-ssl.md
```diff
@@ -28,6 +28,8 @@ Tyk supports TLS connections, and as of version 2.0 all TLS connections will als
...
+
 The `min_version` setting is optional, you can set it to have Tyk only accept connections from TLS V1.0, 1.1 and 1.2 respectively.
```
letzya, line 31: This is relevant to both  tyk.conf and dashboard conf so I dont understand why we have info about both, that only dashboard and then both again.
I would personally read the page as a whole @marksou to confirm it makes sense. Cheers

letzya (changes requested):

tyk-docs/content/basic-config-and-security/security/tls-and-ssl/tls-and-ssl.md
sedkis: Yes thank-you, this was a copy/paste error.  I meant to have include both conf files.

sedkis: I've moved the section to the last paragraph so it's less confusing.  Thank you

marksou (approved):

PR #1120, opened 2020-03-20
Update dashboard-login-okta-tib.md

rewsmith:
Added scopes to profiles.json example


marksou:
Hi @rewsmith Looks ok to me, but perhaps you should get someone like @letzya to review as well?

letzya (approved):

PR #1123, opened 2020-03-23
Add comment to not define more than one group id in a cluster of gate…

letzya:
…ways that share one redis


marksou:
@asoorm Can you review please?

sredxny (approved):
LGTM 👍

PR #1125, opened 2020-03-23
Update 3rd-party-identity-providers.md

rewsmith:
Updated requirements and dependencies to mention TIB based SSO also available for Tyk Multi-Cloud


marksou:
@rewsmith Do we need to mention how they get the feature enabled? Via account manager, etc? Have also formatted the note.

rewsmith:
@marksou That's a good call.  I'll add some appropriate words.

rewsmith:
@marksou Appropriate words now added

marksou (approved):
👍

PR #1130, opened 2020-03-26
made it a bit clearer imo

letzya:


marksou (approved):
Made a couple of amendments.

PR #1141, opened 2020-04-07
adds a page for benchmarking

sedkis:


tyk-docs/content/benchmarks/benchmarks.md
```diff
@@ -0,0 +1,44 @@
...
+
+As an API Gateway introduces an extra hop and is also the single point of entry into the ecosystem, it needs to be super performant.  Tyk was designed to be performant from day one, which is also why it is written in GO.
+
+## TL;DR Benchmark results
+The following table summarizes different performance tests using Amazon EC2 virtual instances.
+This is with Tyk performing authentication, rate limiting, and gathering analytics.
+
+*There is a 0.4 ms delay between upstream and load generator.*
+
+|     Hardware specs      |   Throughput (RPS)  |  99% Latency  |    EC2 Cost    |
```
asoorm: `99% Latency` -> `P99 Latency`

asoorm: I would also provide P95 in here also.

This is also completely unrealistic, where I would assume the machines are consuming 99% of the CPU. So maybe it's worth dropping the RPS, and showing P99 latency at 90% of max RPS?

e.g. t2.micro - provide P99 latency running 2659 RPS.

tyk-docs/content/benchmarks/benchmarks.md
```diff
@@ -0,0 +1,44 @@
...
+|-------------------------|---------------------|---------------|----------------|
+| Amazon EC2 (t2.micro)   | 2,954 req / sec     | 32.9 ms       | $ 0.0116 / hour|
+| Amazon EC2 (t2.medium)  | 4,903 req / sec     | 22.2 ms       | $ 0.0464 / hour|
+| Amazon EC2 (m4.large)   | 4,190 req / sec     | 25.2 ms       | $ 0.10 / hour  |
+| Amazon EC2 (c5.2xlarge) | 23,667 req / sec    | 4.93 ms       | $ 0.34 / hour  |
+| Amazon EC2 (c5.9xlarge) | 69,133 req / sec    | 1.2 ms        | $ 1.53 / hour  |
+
+More information on how to recreate these results yourself:
+*** link to AWS testing blog
```
asoorm: Pending Blog Post

asoorm (changes requested):

tyk-docs/content/benchmarks/benchmarks.md
asoorm: I would also caveat that you deployed Tyk in docker. So note that there is approx 10% performance penalty here.

sedkis: @asoorm  re your comment, are you suggesting having a column that says “max throughput” and then another one that says “P99 latency at 90% throughput” ?

what is the benefit to an end user for reading this ?

sedkis: noted the docker deploy comment and added

asoorm: > @asoorm re your comment, are you suggesting having a column that says “max throughput” and then another one that says “P99 latency at 90% throughput” ?
> 
> what is the benefit to an end user for reading this ?

4 columns:

```
Max Throughput at P95, Max Throughput at P99
90% Throughput at P95, 90% Throughput at P99
```

The way it is currently presented - it will mean different things to different groups of people.

* If latency is a concern, users won't understand that by reducing throughput by a little bit, you can significantly reduce the latency. 
* The P99 value is phenomenally low. That's awesome and something to shout about. But there are far too many peeps that just look at the headline figures and come to the wrong conclusions. e.g they may thikn that our latency is bad without bothering to read full story.
* We don't currently show (via P95) what to expect in a normal, healthy transaction. So the P99, is basically showing worst case scenario, probably during a GC cycle.

asoorm: It feels like we just need to present the full picture. This is what you can expect during normal operation. And during a GC cycle >P99, things might slow down by a bit

tyk-docs/content/benchmarks/benchmarks.md
```diff
@@ -0,0 +1,46 @@
...
+
+> **Note:** EC2 instance cost is variable and documented as of April 6, 2020 
+https://aws.amazon.com/ec2/pricing/on-demand/
+
+> **NOTE:** For these tests, Tyk was deployed in Docker, so add & subtract 10% from throughput and P99 latency, respectively, to account for the decreased overhead if you plan to run from host
+
+> **Note:** The P99 Latency is a metric which means that 99% of all API requests experience a lower latency than the given figure.
```
matthewtanner91, line 37: I would outline something like: Simply put, this means that 99% of the time your latency will be lower than the number used here.

Probably still a better way to phrase it to make it accessible to those who still don't understand the terminology used in the pervious sentence. I think it is important to make sure that people really understand what is meant or they may think that you picked the fastest 1% for the metric which is opposite of the case here.

tyk-docs/content/benchmarks/benchmarks.md
sedkis: Added

tyk-docs/content/benchmarks/benchmarks.md
sedkis: Gotcha thanks

tyk-docs/content/benchmarks/benchmarks.md
sedkis: I understand - thanks.  @nerdydread has actually gotten some good metrics for us that we will add for P99 at 90% max capacity instead of 100%

marksou (approved):
LGTM

sedkis:
@marksou go ahead - also let me know if you prefer we move it to "Planning for production" ? im not sure

PR #1142, opened 2020-04-07
Couple of follow ups for 2.9.4

marksou:


tyk-docs/content/tyk-configuration-reference/tyk-gateway-configuration-options.md
```diff
@@ -671,6 +671,28 @@ By default, we set the `X-Generator` header to `tyk.io` when returning errors. Y
 
 New in v2.9.2, you can now configure the OAuth error status code returned. If not set, it defaults to a 403 error.
 
+### override_messages
+
+New in 2.9.4, you can now override the default error code and or message returned by middleware. The following settings can be overridden:
```
furkansenharputlu: ..The following settings can be....
I am not sure saying `settings` here correct. Maybe, "The following are the message ids that you can use to override messages and error codes" like this. I believe you can make it consistent with all sentence.

marksou: Have now updated

furkansenharputlu (approved):

PR #1151, opened 2020-04-14
Edit page for key deletion as well as expiring.

marksou:


letzya:
@marksou what is the difference? GH doesn't show it. thanks

letzya:
Found it's file name key to token?

letzya:
I'm not sure we should do this change per @buger suggestion to add a footer to explain it

tyk-docs/content/basic-config-and-security/security/Authentication & Authorization/physical-key-expiry.md
```diff
@@ -0,0 +1,42 @@
...
+
+Example: To have keys live in Redis for only 24 hours (and be deleted 24 hours post their creation) set it as follow:
+```{.json}
+"session_lifetime": 86400
+```
+If this is not set, then the default is 0 seconds, which means the key will not be deleted from Redis.
+
+This feature works nicely with JWT or OIDC auth methods since the keys get created in Redis the first time they are in use so you know when it will be removed. Be extra careful in the case of keys created by Tyk (Auth token or JWT with individual secrets) and set a big `session_lifetime` otherwise the user might use the key AFTER it has already been removed from Redis.
+
+### Expiring and deleting tokens at the Global level
```
letzya, line 38: but this still had the word "token"

marksou:
@buger could you have another look at this?

PR #1154, opened 2020-04-16
Update api-definitions.md

rewsmith:
Added notes about being able to set api_id with a call to Update API, but only possible on-premise


marksou (approved):
Couple of edits. Good to go

PR #1184, opened 2020-04-24
Restructure the JWT page to be easier to follow

sedkis:
Starts with an example and moves all the details to the bottom.


marksou (approved):
LGTM

PR #1220, opened 2020-05-14
Do not expose the internal detail of /go/src/plugin-build

alephnull:
goplugins require
- compiler version
- gateway source paths
- plugin source paths
to be identical at build time and run time.


marksou (approved):
👍

letzya (changes requested):
So this is a change starting from 2.9.4? Or 2.9.4.1? Or maybe  this docs work for people that use 2.9.3?
If it works for any previous version then we can merge if not we need to adjust the docs (I tested only 2.9.3 since unfortunately I used the command  line that is on the left side of this PR and it failed on 2.9.4.X)

alephnull:
> So this is a change starting from 2.9.4? Or 2.9.4.1? Or maybe this docs work for people that use 2.9.3?

This would be applicable to 2.9.3. However, there are other bugs plaguing 2.9.3 which are not going to be resolved. Either way, this should be official mount path. Because of the way paths are important to goplugins, I might change the internal paths based on need without impacting docs.

PR #1239, opened 2020-05-20
Fix inline link

elacuesta:
Fixes an error in the link to [`api.h`](https://github.com/TykTechnologies/tyk/blob/master/coprocess/api.h) in https://tyk.io/docs/plugins/rich-plugins/rich-plugins-work/#interoperability.

Screenshot at 2020-05-20 13-42-28


marksou (approved):
👍

PR #1251, opened 2020-05-27
removed key values

madebyguie:
Public keys removed for security matters


alephnull (changes requested):
Rather than leaving it blank, can this be:
```
appId: 'ALGOLIA_APP_ID',
apiKey: 'ALGOLIA_API_KEY'
```
This will make it easy for me to substitute this in the deployment pipeline.

marksou (approved):

PR #1255, opened 2020-05-29
Fix search bugs

madebyguie:


marksou (approved):
Went through the Merge conflicts with @itsmartins.

PR #1288, opened 2020-06-15
Fixed typo in the md syntax

letzya:
this is the typo ** HEADERNAME format:


rewsmith (commented):
Minor grammatical error: Drop `is` in the sentence below:

"Convert the **first letter** in each word of an incoming header is to Capital Case"

should be:

"Convert the **first letter** in each word of an incoming header to Capital Case"

marksou (approved):
👍

PR #1295, opened 2020-06-18
Adding algolia scripts to the repo

### LINKED ISSUE #1292: Incorporate algolia setup into repository
Algolia is used to deliver instant search for the online deployment. The population of the index is done by a couple of Ruby scripts that live only on the Buddy filesystem. 

The code outside git should be checked into git and parameterised so that it can run on any pipeline. Separate Algolia accounts are used for staging and prod.

alephnull:
These scripts are used to build the index that the instanstsearch
widget will use. The ALGOLIA_ADMIN_KEY variable is _not_ the
search-only API key, this should not be exposed.

Required env variables are set on Buddy pipelines.


alephnull:
Please tag me before merging, 
- the scripts need to be removed from BuddyFS before merging
- the path to the script needs to be fixed in the dev and prod buddy pipelines

buger:
Can you moved them to "scripts" folder or similar?

PR #1312, opened 2020-06-24
Add footnotes shortcode and replace blacklist/whitelist

buger:
Added a new shortcode to specify the footnotes, which can be used to replace some terms, while still keeping the benefits of SEO. Terms should be defined in advance inside "content/terms.yaml" file. For example, the current file looks like this:
```
"terms":
  "allowlist": "Allowlist - explicitly allowing access to some identified entities. Also commonly known as a whitelist."
  "blocklist": "Blocklist - explicitly blocking access to some identified entities. Also commonly known as a blacklist."
```

Fairly easy to use, just wrap the term where you want to add a footnote to the following shortcode: `{{<fn>}}whitelist{{</fn>}}`. When done, at the bottom of the page, you automatically see the footnote, and it will be cross-linked. Example:
image

In some of the cases, the word you have, not directly maps to the term, for example, "allowlisting" is not the same as "allowlist". Shortcode has support for dealing with it, by specifying term via argument, like this: `{{<fn allowlist>}}allowlisting{{</fn>}}`. Additionally, in some of the cases, when you have a link, and you can't wrap it to the shortcode, you can just put the empty shortcode `{{<fn allowlist>}}{{</fn>}}`

This PR replace all usage of blacklist and whitelist in our docs, in favor of blocklist and allowlist, and in a lot of the cases in favor of verbs. E.g. instead of "Adding to blocklist", just "block". 

A lot of the parts became clearer. 
For sure there are some things which should be renamed on the application level, and we need to plan it for 3.1 or similar.


marksou (approved):
👍

buger:
Short coded part was merged long ago. The PR itself should be revisited after docs re-design.

marksou:
Closing as addressed in #1879

PR #1319, opened 2020-06-26
rewrite Plugin docs

sedkis:
The new content to be approved, everything else is just moving stuff around:
1) request plugins
image

2) supported auth languages
image


tyk-docs/content/plugins/request-plugins.md
```diff
@@ -0,0 +1,55 @@
...
+|-------------------------|----------|-------------|-----------|-----------|
+| Can modify the Header   | ✅       | ✅          | ✅       | ✅  
+| Can modify the Body     | ✅       | ✅          | ✅       |✅
+| Can modify Query Params | ✅       | ✅          | ✅       |✅
+| Can view Token Details  |          | ✅          |✅          |✅
+| Can modify Token (metadata, quota, context-vars, tags, etc)|          | ✅          |          |
+
+
+### Return Overrides / ReturnOverrides  
+You can have your plugin finish the request lifecycle and return a response to the requestor.
```
letzya: to clarify I'd add that you modify the response body.

tyk-docs/content/plugins/response-plugins.md
```diff
@@ -0,0 +1,47 @@
...
+menu:
+  main:
+    parent: "Custom Plugins"
+weight: 20
+---
+
+Since Tyk 3.0 we have incorporated response hooks, this type of hook allows you to modify the response object before its being sent to the client. The flow is follows:
+
+- Tyk receives the request.
+- Tyk runs the full middleware chain, including any other rich plugins hooks like Pre, Post, Custom Authentication, etc.
```
letzya: maybe remove the word "rich"? since it applies to any plugin - go, jsvm etc.

tyk-docs/content/plugins/response-plugins.md
```diff
@@ -0,0 +1,47 @@
+---
+date: 2017-03-24T15:45:13Z
+title: Response Plugins
+menu:
+  main:
+    parent: "Custom Plugins"
+weight: 20
+---
+
+Since Tyk 3.0 we have incorporated response hooks, this type of hook allows you to modify the response object before its being sent to the client. The flow is follows:
```
letzya: can you add a clarification that this called after the gw received a response from the upstream? unlike return override that never gets to the backend

tyk-docs/content/plugins/request-plugins.md
```diff
@@ -0,0 +1,55 @@
+---
+date: 2017-03-24T15:45:13Z
+title: Request Plugins
+menu:
+  main:
+    parent: "Custom Plugins"
+weight: 10
+---
+
+There are 3 different phases in the [request lifecycle](/docs/getting-started/key-concepts/middleware-execution-order/) you can inject custom plugins, 4 if you include [Authentication plugins](/docs/plugins/auth-plugins/).  There are performance advantages to picking the correct phase, and of course that depends on your use case and what functionality you need.
```
letzya: I personally would not say 3 and then 4, just 4 since auth hook is like any other hook.

Question - which "performance advantages to picking the correct phase,"?

tyk-docs/content/plugins/request-plugins.md
```diff
@@ -0,0 +1,55 @@
...
+
+### Hook Capabilities
+| Functionality           |   Pre    |  Auth       | Post-Auth |    Post   |
+|-------------------------|----------|-------------|-----------|-----------|
+| Can modify the Header   | ✅       | ✅          | ✅       | ✅  
+| Can modify the Body     | ✅       | ✅          | ✅       |✅
+| Can modify Query Params | ✅       | ✅          | ✅       |✅
+| Can view Token Details  |          | ✅          |✅          |✅
+| Can modify Token (metadata, quota, context-vars, tags, etc)|          | ✅          |          |
```
letzya: I still think it's confusing. maybe write "Can modify user session" - a session is a common term

context-vars - this is not part of the user session but the other way around. you can access the context using the request object and from the request, you get everything - session object, context-var, headers.
check https://site-dev.tykbeta.com/docs/plugins/rich-plugins/rich-plugins-data-structures/#sessionstate-session_stateproto

letzya (changes requested):

tyk-docs/content/plugins/request-plugins.md
sedkis: I'm updating this!

tyk-docs/content/plugins/response-plugins.md
sedkis: Done:
`
Since Tyk 3.0 we have incorporated response hooks, this type of hook allows you to modify the response object sent by the upstream. The flow is follows:
`

tyk-docs/content/plugins/request-plugins.md
sedkis: Done

sedkis: >Question - which "performance advantages to picking the correct phase,"?

The ones we spoke about on Slack!

tyk-docs/content/plugins/request-plugins.md
sedkis: Done!

marksou (approved):
👍

tyk-docs/content/plugins/request-plugins.md
letzya: I don't remember but you mean to fail sooner than later but regardless we should explain it here not just stating it vaguely.

PR #1323, opened 2020-06-30
Create include file

### LINKED ISSUE #1321: duplicated content for OIDC flow
## Relevant Document

https://site-dev.tykbeta.com/docs/basic-config-and-security/security/authentication-authorization/openid-connect/
https://site-dev.tykbeta.com/docs/advanced-configuration/integrate/api-auth-mode/open-id-connect/

## Describe the error

There seem to be 2 versions of oidc flow - under basic configuration and under advanced (one of them with 9 steps the other 10). To avoid users' confusion, we should avoid content duplication and have one version in one place (and also decide if it's basic or advanced :) )

marksou:
Fixes #1321 Created an include file so both locations have the same content, but edited only in one place.

### DIFF
--- a/tyk-docs/content/advanced-configuration/integrate/api-auth-mode/open-id-connect.md
+++ b/tyk-docs/content/advanced-configuration/integrate/api-auth-mode/open-id-connect.md
@@ -7,21 +7,7 @@ menu:
 weight: 0
 ---
 
-Tyk comes with support for OpenID Connect (OIDC) Identity Tokens provided by any standards compliant OIDC provider.
-
-### The OIDC Flow:
-
-1.  User requests access to resource via a supported OIDC Provider (e.g. Google).
-2.  User logs into resource provider and grants scope access to their data.
-3.  Identity Provider generates OAuth token set and OIDC ID Token. ID Token is signed by provider with their public key.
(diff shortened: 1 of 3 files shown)


letzya (changes requested):
@marksou sorry but I disagree with this method since I feel it's confusing from UX perspective. My journey: As a user, I want to read all docs about oidc but when I see a duplicated content I start comparing to understand the difference especially when I'm stuck on something and need answers. So this fix will help tyk not to make mistakes but as a user, I'd still waste time reading both just to make sure I'm not missing anything.
Also, it doesn't make sense to have the same topic as basic and advanced.
Why can't we use a hyperlink instead, the way things are done on the web?

marksou:
@letzya My argument is by linking to another page, a user is less likely to click on it as they lose where they were in the first place which is not good UX. If you are happy for us to lose the page in the Basic section, I'm happy to do that.

letzya:
We need to understand what we want to write. User will not less likely if he is after a piece of information that is clearly displayed and explained. I think we should never use include for content. In this case we can write a short comment in the basic conf page and add a link

buger:
Closing since issue was solved with recent docs restructure

PR #1324, opened 2020-06-30
Fixed tab, fn and include shortcodes, migration from mmark

buger:
## Tab changes

In order to use tabs you now should use the following structure:
```
{{< tabs_start >}}

  {{< tab_start "Cloud" >}}
   ....
  {{< tab_end >}}

{{< tabs_end >}}
```

Note that there are no "closing" tags with slash symbols like previously (e.g. `{{</tabs>}}`).

## Shared content

All the "shared" content now should live inside "content/shared" folder, like a regular markdown page.
And be included like this (by file name without .md file extension).

```
{{< include "create_api" >}}
```

## Footnotes changes

Previously footnotes was specified inside the `./content/terms.yaml` and now it is inside `./data/terms.yaml`.
It should fix the production deployment issue as well.

## mmark migration

See `try-out-tyk/tutorials/create-api.md` as example


marksou (approved):
👍

PR #1348, opened 2020-07-13
current link leads to openid struct.

letzya:
But this will need to be updated every time so maybe we should put it in the docs not link to the code


marksou (approved):

buger:
@letzya what is the actual change it just shows that some empty lines was updated?

letzya:
It's from july 2020, I guess it was fixed by now so no diff

PR #1364, opened 2020-07-22
Add alias for previous URL

marksou:
@buger I have set the alias but it didn't seem to work when testing locally. Should we merge and test on staging?

### DIFF
--- a/tyk-docs/content/planning-for-production/redis-mongodb-sizing.md
+++ b/tyk-docs/content/planning-for-production/redis-mongodb-sizing.md
@@ -3,7 +3,9 @@ title: Redis and Mongo Sizing
 menu:
   main:
     parent: "Redis and MongoDB"
-weight: 2 
+weight: 2
+aliases:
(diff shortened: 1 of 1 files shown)


tyk-docs/content/planning-for-production/redis-mongodb-sizing.md
```diff
@@ -3,7 +3,9 @@ title: Redis and Mongo Sizing
 menu:
   main:
     parent: "Redis and MongoDB"
-weight: 2 
+weight: 2
+aliases:
+  - /docs/analytics-and-reporting/redis-mongodb-sizing/#a-name-mongodb-a-mongodb
```
buger: Remove "#" part from the URL

marksou: Done.

marksou:
I've tested this change locally and the alias now redirects ok

PR #1374, opened 2020-08-03
Update tyk-sync.md

rewsmith:
Updated to include docker instructions, similar to the equivalent GH repo


marksou (approved):
👍  After adding some links, etc

sedkis (approved):

PR #1385, opened 2020-08-19
redis key events are now true by default. Merge when 3.0.1 is released

sredxny:
In Dashboard 3.0.1 the redis key events will be propagated no matter what configuration we have in mongo for the organization, this is the documentation for https://github.com/TykTechnologies/tyk-analytics/pull/1900

This should be merge when 3.0.1 is released


sredxny:
Documentation: https://github.com/TykTechnologies/tyk-docs/pull/1385

marksou (approved):
👍

PR #1388, opened 2020-08-25
Copy update and glossary link update

emma-m-campbell:
I updated the glossary link as well but not sure if it's the correct one or not?


marksou (approved):
👍

PR #1392, opened 2020-08-25
Title and copy changes,removed success note

emma-m-campbell:
Just confirmed with Artem that we are now allowing more than one control plane per environment so the messaging is no longer needed.


marksou (approved):
👍

PR #1393, opened 2020-08-25
Removed prerequisites and copy changes

emma-m-campbell:
We were missing the step about needing to manually deploy the Edge Gateway - have added that in and moved the prerequisite text into that step.


marksou (approved):
👍

PR #1396, opened 2020-08-25
Minor copy change

emma-m-campbell:
Removed the bit about adding people to a team as that is an optional step so they may or may not have completed it.


marksou (approved):
👍

PR #1413, opened 2020-08-28
change health check message return, updated for V3.0

Lvceo:
Now the health check message is following the https://inadarei.github.io/rfc-healthcheck/


sedkis (approved):

sedkis:
Might be a good idea to find any instances of "hello Tiki" in our docs and update them with the new Hello message

PR #1428, opened 2020-09-09
TT-41 removed not supported providers for tib

sredxny:
The documentation currently shows some providers that are not really supported by TIB. In this PR we update that info


marksou (approved):
👍

PR #1435, opened 2020-09-15
Authorise endpoint API name clarification

joshblakeley:
The Authorise client endpoint on the dashboard builds the URL to call the Gateway in the form: 

`config.Global().TykApi.Host + ":" + config.Global().TykApi.Port + "/" + apiID + "/tyk/oauth/authorize-client/"`

Where the APIID is read from the path of the dashboard endpoint. 

However the gateway API builds its path in the form:

`spec.Proxy.ListenPath + pathSeparator + "tyk/oauth/authorize-client{_:/?}"`

Meaning the dashboard should in fact pass the API name and not the ID, otherwise a 404 will be received.

### DIFF
--- a/tyk-docs/content/tyk-apis/tyk-dashboard-api/oauth-key-management.md
+++ b/tyk-docs/content/tyk-apis/tyk-dashboard-api/oauth-key-management.md
@@ -147,7 +147,7 @@ curl -vX DELETE -H "Authorization: {{API Access Credentials}}" \
 | Type         | Form-Encoded                                   |
 | Body         | Fields (see below)                             |
 
-* `api_id`: Unlike the other requests on this page, this should be the `api_id` value and **NOT** the API's `id` value. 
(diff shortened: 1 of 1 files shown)


marksou (approved):
👍

letzya (approved):

letzya (changes requested):
@joshblakeley about `Unlike the other requests on this page, this should be the` - shouldn't it be api name everywhere? maybe if it's on the old cloud it needs id?

buger:
The way how it is currently described in the docs is the way how it should work. 
So I would rather treat it as a product bug. 
This API is actually using listen path as argument not the API_ID, but in context of our legacy cloud it is API ID (since internally listen path is API ID, and slug is using durign some weird nginx rewriting).

 This functionality can be considered broken on on-prem and Ara, it should use api_id instead. 

Create JIRA ticket https://tyktech.atlassian.net/browse/TT-1727

PR #1463, opened 2020-10-14
adds note on how to make uptime tests

sedkis:
@gernest added you as you did some work on the uptime tests. can you confirm the details added are true please


marksou (approved):
👍

PR #1465, opened 2020-10-16
TT-423 update grpc doc, as now supports h2c

sredxny:
Documentation for https://tyktech.atlassian.net/browse/TT-423 , as h2c was added as a new feature,

### DIFF
--- a/tyk-docs/content/getting-started/key-concepts/grpc-proxy.md
+++ b/tyk-docs/content/getting-started/key-concepts/grpc-proxy.md
@@ -12,7 +12,7 @@ url: "/key-concepts/grpc-proxy"
 
 Tyk supports gRPC passthrough proxying when using HTTP/2 as a transport (the most common way to deploy gRPC services). The only requirement is enabling HTTP/2 support on the Gateway side, for both incoming and upstream connections, by setting `http_server_options.enable_http2` to true in your Gateway config file.
 However, in the scenario that you want to secure the connection between 2 services that call each other then you can use h2c (that is the non-TLS version of HTTP/2), this can be enabled at api level by setting `h2c` as protocol in the address of the grpc server (Example: `h2c://mygrpcserver.com`).
-You also need to set your `listen_path` in your API definitions and a specific port where the service will be exposed.
+You also need to set your `listen_path` in your API definition and a specific port where the service will be exposed.
 
 The gRPC over HTTP2 specification defines the rules on how the gRPC protocol maps to a HTTP request. In the context of the API Gateway, we are interested in the following:
 
--- a/tyk-docs/content/getting-started/key-concepts/grpc-proxy.md
+++ b/tyk-docs/content/getting-started/key-concepts/grpc-proxy.md
@@ -10,15 +10,15 @@ url: "/key-concepts/grpc-proxy"
 
 ### Using Tyk as a gRPC Proxy
 
-Tyk supports gRPC passthrough proxying when using HTTP/2 as a transport (the most common way to deploy gRPC services). The only requirement is enabling HTTP/2 support on the Gateway side, for both incoming and upstream connections, by setting `http_server_options.enable_http2` and `proxy_enable_http2` to true in your Gateway config file.
-However, in the scenario that you want to secure the connection between 2 services that call each other then you can use h2c (that is the non-TLS version of HTTP/2), this can be done by enabling `proxy_enable_h2c` in the gateway config file or more granular level enabling `proxy.enable_h2c` in the API definition.
-You also need to set your `listen_path` in your API definitions.
+Tyk supports gRPC passthrough proxying when using HTTP/2 as a transport (the most common way to deploy gRPC services). The only requirement is enabling HTTP/2 support on the Gateway side, for both incoming and upstream connections, by setting `http_server_options.enable_http2` to true in your Gateway config file.
+However, in the scenario that you want to secure the connection between 2 services that call each other then you can use h2c (that is the non-TLS version of HTTP/2), this can be enabled at api level by setting `h2c` as protocol in the address of the grpc server (Example: `h2c://mygrpcserver.com`).
+You also need to set your `listen_path` in your API definitions and a specific port where the service will be exposed.
 
-The gRPC over HTTP2 specification defines the rules on how the gRPC protocol maps to a HTTP request. In the context of the API Gateway, we are intersted in the following:
+The gRPC over HTTP2 specification defines the rules on how the gRPC protocol maps to a HTTP request. In the context of the API Gateway, we are interested in the following:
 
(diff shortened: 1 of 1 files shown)


joshblakeley:
@sredxny As mentioned on slack - the gRPC proxying docs need to have information for users on how to use:

- Secure gRPC proxy
- Insecure proxying by leveraging h2c proto
- gRPC streaming - how to set it up and WHY we use the settings we recommend (get info from @matiasinsaurralde if unsure)
- What data is in a gRPC request that comes through Tyk so we know how to access it from middleswares etc


Also is this now debunked or needs rewording? ```- HTTP path follows the format `/{Service-Name}/{method name}`, for example: `/google.pubsub.v2.PublisherService/CreateTopic`. ```

sredxny:
@joshblakeley 2 things:

- Why do you think this needs rewording? Am trying to understand what could be confusing or not clear so I can make the proper rewording: `- HTTP path follows the format `/{Service-Name}/{method name}`, for example: `/google.pubsub.v2.PublisherService/CreateTopic`.`
- About the information that comes into tyk, do you think that it's answered with https://github.com/TykTechnologies/tyk-docs/pull/1465/files#diff-93ae6962eb18f89a23d0bb73e5db83e6c5625f779c0e8e81b16e35de2269bc93R21 ?

Am seeing that we're adding some subtitles with only one line of content. Wouldn't be better to consolidate it? anyway am thinking how

tyk-docs/content/getting-started/key-concepts/grpc-proxy.md
```diff
@@ -10,16 +10,28 @@ url: "/key-concepts/grpc-proxy"
...
+- Set `disable_ports_whitelist` to true, so the gateway can use additional ports to expose the service.
+
+#### Secure gRPC Proxy
+Tyk Supports secure gRPC proxy connections, in order to do so you only need to attach a certificate to the API that you want to expose just as you do for regular apis, after that you can consume the service via https.
+
+#### Insecure gRPC Proxy (H2C)
+For those scenarios that you want to connect 2 services that call each other or just ned an insecure connection you can use h2c (that is the non-TLS version of HTTP/2). Tyk supports h2c, this can be enabled at api level by setting `h2c` as protocol in the address of the grpc server (`target_url`) e.g: `h2c://mygrpcserver.com`.
 
+#### About streaming
+For gRPC streaming is recommended to set a low value in the gateway config `http_server_options.flush_interval` (this value is in milliseconds)
```
joshblakeley: Can we state why this is required

tyk-docs/content/getting-started/key-concepts/grpc-proxy.md
```diff
@@ -10,16 +10,28 @@ url: "/key-concepts/grpc-proxy"
 
 ### Using Tyk as a gRPC Proxy
 
-Tyk supports gRPC passthrough proxying when using HTTP/2 as a transport (the most common way to deploy gRPC services). The only requirement is enabling HTTP/2 support on the Gateway side, for both incoming and upstream connections, by setting `http_server_options.enable_http2` and `proxy_enable_http2` to true in your Gateway config file.
+Tyk supports gRPC passthrough proxying when using HTTP/2 as a transport (the most common way to deploy gRPC services).
+However,
```
joshblakeley: No need to caveat with `however`. Simply explain how it should be configured should be enough

tyk-docs/content/getting-started/key-concepts/grpc-proxy.md
```diff
@@ -10,16 +10,28 @@ url: "/key-concepts/grpc-proxy"
...
 
-You also need to set your `listen_path` in your API definitions.
-
-The gRPC over HTTP2 specification defines the rules on how the gRPC protocol maps to a HTTP request. In the context of the API Gateway, we are intersted in the following:
+The gRPC over HTTP2 specification defines the rules on how the gRPC protocol maps to a HTTP request. In the context of the API Gateway, we are interested in the following:
 
 - HTTP path follows the format `/{Service-Name}/{method name}`, for example: `/google.pubsub.v2.PublisherService/CreateTopic`. You can use this feature to apply standard ACL rules via Keys and Policies, or use URL rewrite plugins in our [Endpoint Desiger](/docs/transform-traffic/url-rewriting/#a-name-url-rewrite-with-endpoint-designer-a-rewrite-a-url-with-the-endpoint-designer).
 - HTTP method is always `POST`.
-gRPC custom request metadata is added as HTTP headers, where metadata key is directly mapped to the HTTP header with the same name.
+gRPC custom request metadata is added as HTTP headers, where metadata key is directly mapped to the HTTP header with the same name.
```
joshblakeley, line 20: Can we link to the gRPC spec page on it's HTTP2 mapping here so people can get more info if they need. https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md

tyk-docs/content/getting-started/key-concepts/grpc-proxy.md
joshblakeley: And also a good value to set specifically

tyk-docs/content/getting-started/key-concepts/grpc-proxy.md
```diff
@@ -10,16 +10,28 @@ url: "/key-concepts/grpc-proxy"
...
+- Enable  HTTP/2 support on the Gateway side, for both incoming and upstream connections, by setting `http_server_options.enable_http2` to true in your Gateway config file.
+- Set `disable_ports_whitelist` to true, so the gateway can use additional ports to expose the service.
+
+#### Secure gRPC Proxy
+Tyk Supports secure gRPC proxy connections, in order to do so you only need to attach a certificate to the API that you want to expose just as you do for regular apis, after that you can consume the service via https.
+
+#### Insecure gRPC Proxy (H2C)
+For those scenarios that you want to connect 2 services that call each other or just ned an insecure connection you can use h2c (that is the non-TLS version of HTTP/2). Tyk supports h2c, this can be enabled at api level by setting `h2c` as protocol in the address of the grpc server (`target_url`) e.g: `h2c://mygrpcserver.com`.
 
+#### About streaming
```
joshblakeley: Retitle to `gRPC Streaming`

joshblakeley (commented):
Im happy not consolidating it because we should make each section more informative over time. Lets not treat this as "completed docs" because it isnt. This is just the minimum we need to get a user going but it can be MUCH better with code snippets - example client and server code / curl commands etc.

Regarding the path information it is irrelevant because the API must listen on `/`. It would be more correct to say that when targeting an upstream gRPC server from within Tyk the HTTP path can be used to targret specific functions inside middlewares.

tyk-docs/content/getting-started/key-concepts/grpc-proxy.md
matiasinsaurralde: We can mention something like:
```
A low flush interval is required in order to forward data to the downstream as soon as the upstream replies. A high flush interval would delay the communication. We recommend the lowest possible value: 1 (1 millisecond).
```

joshblakeley:
Looking really nice @sredxny - can we get another dev from Tyk to try it out using these docs as a guide and see how they get along. That should help us know for sure the docs are instructive! Happy to merge anyway in meantime as we can iterate 🥇

joshblakeley (approved):

matiasinsaurralde (approved):

PR #1466, opened 2020-10-19
Custom keys

marksou:


tyk-docs/content/advanced-configuration/manage-multiple-environments/move-keys-between-environments.md
```diff
@@ -17,4 +17,4 @@ In order to use a legacy key in a new environment, simply extract the key from t
 
 ### Creating Custom Keys
 
-To create a key with a custom identifier, ie Token, simply create a new key by making a `PUT` to the `/tyk/keys/{keyId}`, where `keyId` is your custom value.
\ No newline at end of file
+To create a key with a custom identifier, ie Token, simply create a new key by making a `POST` to the `/tyk/keys/{keyId}`, where `keyId` is your custom value.
```
sedkis: There is one more on this page, the sub-heading right above this needs to also be changed to `POST`

tyk-docs/static/others/gateway-swagger.yaml
```diff
@@ -596,7 +596,7 @@ paths:
                 items:
                   type: string
     post:
-      summary: Create a key
+      summary: Create a key or add custom Key
```
sedkis: This is incorrect, this needs to be moved down to a child of the next sub-heading `'/tyk/keys/{keyID}':`.  which means it becomes a sibling to the other below that was changed

sedkis (changes requested):

PR #1468, opened 2020-10-21
Design changes

marksou:
Design changes based on Figma file. Need to look at Note/Warning icon placement


buger:
I think font size here should be increased by one pixel too:
image

buger:
For the Note icon, try using "after" selector like this:

```
.note strong::after {
  content: url('../images/small_triangle.png');
}
```

marksou:
Have added the icons. The only thing left is the code blocks. I can't seem to remove the border that appears on the right and bottom.
Screenshot 2020-10-21 at 16 09 52

buger (approved):

PR #1512, opened 2020-12-15
Clarify Tyk Cloud custom domain usage

excieve:
Tyk Cloud control planes and edges have different kinds of ingresses. It's important to CNAME a correct one in each type of a deployment.


marksou (approved):
👍

PR #1516, opened 2020-12-21
add looping example

sedkis:
Adds looping example.

image

It's a child of "url rewriting"


marksou (approved):
LGTM

tyk-docs/content/advanced-configuration/transform-traffic/looping.md
```diff
@@ -0,0 +1,100 @@
...
+    parent: "URL Rewriting"
+weight: 5 
+---
+
+## Overview
+
+Instead of rewriting to a HTTP endpoint using [URL Rewriting](../url-rewriting), you now can tell Tyk to internally run its request pipeline one more time, but for another specified endpoint. This is called <b>looping</b>.
+
+In order to specify a loop, in the target URL you specify a string in the following format: 
+
```
letzya, line 15: @sedkis maybe clarify that this redirects to another endpoint of the current api

tyk-docs/content/advanced-configuration/transform-traffic/looping.md
```diff
@@ -0,0 +1,100 @@
...
+For example:
+
+```
+tyk://123/myendpoint?method=GET
+```
+
+### Loop Limiting
+
+You can tell Tyk to limit the number of loops by adding the following flag
```
letzya: This is to check RL and Quota

tyk-docs/content/advanced-configuration/transform-traffic/looping.md
```diff
@@ -0,0 +1,100 @@
...
+```
+
+You can also loop to another API as by specifying the API name or id instead of self: 
+```
+tyk://<API_ID>/<path>.
+```
+
+Combined with our advanced URL rewriter rules, it can be turned into a powerful logical block, replacing the need for writing middleware or virtual endpoints in many cases.
+
+This is very performant because Tyk will not do another network call when detecting a loop.
```
letzya: I think we should add that there's a default limit level in the request level (i.e. set for every request) to avoid endless recursion 
`defaultLoopLevelLimit = 5` and that you can overwrite it (as shown further down)
and that the user will get an error if it's higher `"Loop level too deep. Found more than %d loops in single request", limit)`
Also, you can set the limit loop level only once per request. After that, you can't redirect again with a new loop level limit

sedkis: Can you edit the doc and add this?  I want to be sure we have it accurate @letzya

tyk-docs/content/advanced-configuration/transform-traffic/looping.md
sedkis: i thought `?check_limits=true` is for RL and quota, and `loop_limit` is for loop limit

tyk-docs/content/advanced-configuration/transform-traffic/looping.md
sedkis: yes good idea

tyk-docs/content/advanced-configuration/transform-traffic/looping.md
letzya: check_limits=true is for RL and quota
loop_limit - limit the level of the loop

letzya (approved):

PR #1524, opened 2021-01-15
Update rich-plugins-work.md

jonathanbernal25:
Re-arranging the middleware list in terms or order of execution, from top to bottom. Also relabeling "CustomAuthCheck" to "AuthCheck" in order to match the standard hook name in the API definition.


matiasinsaurralde (approved):
LGTM

marksou (approved):
👍

PR #1530, opened 2021-01-22
remove tyk cli reference and replace with dockerized tyk-gateway command

sedkis:
In this Cloud version of a Plugin deployment example, we explain to users that they should use the Tyk-CLI tool in order to build a bundle.  The tyk-cli tool hasn't been updated in over 4 years.

Furthermore, it's not Dockerized and we can't expect end users to download Go in order to use a simple bundler.

This change will just use a Gateway image to bundle the plugin using the built-in "bundle" command


marksou (approved):
👍

PR #1541, opened 2021-02-03
Oauth2 update

marksou:


tyk-docs/content/basic-config-and-security/security/Authentication & Authorization/oauth-2-0.md
```diff
@@ -53,7 +53,7 @@ This seems like a complicated process and very verbose - however in actuality, t
 - Step (4) Creating a page to receive the OAuth POST request, log the user in, authorise the client ID and redirect them back to the client app
 - Step (9) Create a webhook endpoint that accepts a POST request in order to store and update OAuth key data
```
rewsmith, line 64: We should update step 5 to reference the Dashboard API rather than the Gateway API, so instead of:

"If the user accepts the Client access and has authenticated successfully, your app calls the Tyk REST API OAuth Authorization endpoint (/tyk/oauth/authorize-client/) with the POST parameters that the requesting client sent"

instead use:

"If the user accepts the Client access and has authenticated successfully, your app calls the Tyk Dashboard OAuth API Authorization endpoint (/authorize-client/) with the POST parameters that the requesting client sent"

Also add a note that Gateway API also available, as follows:

* The Tyk Gateway also exposes an equivalent REST API authorization endpoint (/tyk/oauth/authorize-client/).  In some scenarios, for example where access to the Dashboard API from the authentication server may be restricted, the Gateway API could be used instead.

rewsmith (changes requested):
We should also update the Authorization Code Grant Type page with the same note about there being an equivalent Tyk Gateway REST API ...

[Request an authorization code](https://tyk.io/docs/basic-config-and-security/security/authentication-authorization/oauth2.0/auth-code-grant#request-an-authorization-code
)

Add the note to URL: Uses the Dashboard API client authorization endpoint /authorize-client/.

tyk-docs/content/basic-config-and-security/security/Authentication & Authorization/oauth-2-0.md
marksou: Changed

PR #1545, opened 2021-02-04
Update oauth-key-management.md

sedkis:
The method is actually a `PUT`, there is no resource for `POST` at that endpoint


marksou (approved):
👍

rewsmith (commented):
I'm pretty sure this is a POST endpoint, not a PUT

PR #1549, opened 2021-02-10
Restructure navigation

sedkis:
Restructuring the navigation to have an emphasis and focus on getting started


marksou (approved):
Go, go, go!

PR #1559, opened 2021-02-17
add documentation for migration guide

jay-deshmukh:
- FIx https://tyktech.atlassian.net/browse/TT-1621

### DIFF
--- a/tyk-docs/content/graphql/migration-guide.md
+++ b/tyk-docs/content/graphql/migration-guide.md
@@ -19,12 +19,13 @@ Old Api definitions will continue to work for the gateway
 
 - Data Source kind are `REST` or `GraphQL` regardless of api being internal or not.
 
-- Incase of internal apis that are accessed via `tyk://`scheme, `graphql.engine.data_sources[n].internal` property is set to true.
+- In case of internal apis that are accessed via `tyk://`scheme, `graphql.engine.data_sources[n].internal` property is set to true.
 
 - Each dataSources needs to be defined with a unique name `graphql.engine.data_sources[n].name`.
 
 - Each field connected to the data source is expected to be configured for mapping under `graphql.engine.field_configs` regardless of it requiring mapping or not.
 
+- It is important that all new graphql apis have version `graphql.version` property set to `2`.
 
 Examples
 
@@ -220,7 +221,6 @@ Examples
       }
     ]
   },
-  "type_field_configurations": [],
   "execution_mode": "executionEngine",
   "version": "2",
   "playground": {
--- a/tyk-docs/content/graphql/migration-guide.md
+++ b/tyk-docs/content/graphql/migration-guide.md
@@ -1,0 +1,228 @@
+---
+title: "Migrating to 3.2"
+date: 2020-06-03
+menu:
+  main:
+    parent: "GraphQL"
+weight: 1
+aliases:
+    - /graphql/migration/
+---
+
+As of 3.2 graphql schema under tyk api definition (i.e `api_definition.graphql`) changed significantly, hence graphql api definitions created in previous beta versions are not supported via UI and need to go through following changes manually.
+
+- To improve performance now a single Data Source can be used to link to multiple fields instead of having an independent data source for every field hence `graphql.type_field_configurations` is now obsolete and new data sources can be defined under `graphql.engine.data_sources` (see example below).
+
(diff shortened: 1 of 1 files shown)


marksou (approved):

vaske (approved):

tyk-docs/content/graphql/migration-guide.md
```diff
@@ -0,0 +1,228 @@
...
+
+- To improve performance now a single Data Source can be used to link to multiple fields instead of having an independent data source for every field hence `graphql.type_field_configurations` is now obsolete and new data sources can be defined under `graphql.engine.data_sources` (see example below).
+
+- Data Source kind are `REST` or `GraphQL` regardless of api being internal or not.
+
+- Incase of internal apis `graphql.engine.data_sources[n].internal` property is set to true.
+
+- Each dataSources needs to be defined with a unique name `graphql.engine.data_sources[n].name`.
+
+- Each field connected to the data source is expected to be configured for mapping under `graphql.engine.field_configs`
```
devsergiy: @jay-deshmukh should we say here that the field `should be` configured nevertheless it is requires mapping or not?

tyk-docs/content/graphql/migration-guide.md
```diff
@@ -0,0 +1,228 @@
...
+date: 2020-06-03
+menu:
+  main:
+    parent: "GraphQL"
+weight: 1
+aliases:
+    - /graphql/migration/
+---
+
+As of 3.2 graphql schema under tyk api definition (i.e `api_definition.graphql`) changed significantly, hence graphql api definitions created in previous beta versions are not supported via UI and need to go through following changes manually.
```
pvormste, line 12: Maybe we should mention explicitly that old api definitions will continue to work for the gateway

tyk-docs/content/graphql/migration-guide.md
```diff
@@ -0,0 +1,228 @@
...
+
+- To improve performance now a single Data Source can be used to link to multiple fields instead of having an independent data source for every field hence `graphql.type_field_configurations` is now obsolete and new data sources can be defined under `graphql.engine.data_sources` (see example below).
+
+- Data Source kind are `REST` or `GraphQL` regardless of api being internal or not.
+
+- Incase of internal apis `graphql.engine.data_sources[n].internal` property is set to true.
```
pvormste: do we need to explain, that internal APIs are the ones that are accessed via `tyk://` scheme?

emma-m-campbell:
@jay-deshmukh From a user perspective, we're looking at directing people from the dashboard to this docs page so that they can make the necessary changes. Could be just my lack of technical knowledge but I'm not sure that this describes enough what the user should do. It describes what changes have happened and why but not so much specifically what I need to change in the raw definition. Maybe it could use wording like "you need to remove this and replace it with this, or edit this to this" and then once 'x' is done then it will be unbroken and you can use the dashboard again? Could we get a Tyk developer from outside the manage squad to read this and see if they understand what needs to be done to unbreak the changes?

jay-deshmukh:
@ConsM @ifrim can you take a look at this with your fresh eyes please, and confirm if it's clear how to redefine schema. 
Thanks

ConsM:
From my little knowledge, guidance around data sources look good.
Maybe we need a similar approach for old and new graphql definitions ?

tyk-docs/content/graphql/migration-guide.md
```diff
@@ -0,0 +1,232 @@
...
+
+**Note**  
+Old Api definitions will continue to work for the gateway
+
+
+- To improve performance now a single Data Source can be used to link to multiple fields instead of having an independent data source for every field hence `graphql.type_field_configurations` is now obsolete and new data sources can be defined under `graphql.engine.data_sources` (see example below).
+
+- Data Source kind are `REST` or `GraphQL` regardless of api being internal or not.
+
+- Incase of internal apis that are accessed via `tyk://`scheme, `graphql.engine.data_sources[n].internal` property is set to true.
```
ifrim: I think a space is missing here.
`In case` instead of `Incase`.

ifrim:
@jay-deshmukh After reading this I think that I would be confident enough that I would know what needs to be changed.

jay-deshmukh:
thanks @ifrim and @ConsM 🙏

marksou:
Merged

PR #1564, opened 2021-02-22
TT- 256 TT-959 DCR Docs

gregdelhon:
First version of the DCR docs. Needs more work and polishing.

### DIFF
--- a/tyk-docs/content/tyk-developer-portal/dynamic-client-registration.md
+++ b/tyk-docs/content/tyk-developer-portal/dynamic-client-registration.md
@@ -1,0 +1,39 @@
+---
+date: 2017-03-24T17:10:33Z
+title: Dynamic Client Registration
+menu:
+  main:
+    parent: "Tyk Developer Portal"
+weight: 1
+---
+
+## OAuth 2.0 Dynamic Client Registration Protocol (DCR)
+Available on version 3.2.0 onwards - currently not generally available but if you want beta access, get in touch! 
+
+# What is Dynamic client registration? 
+
+DCR is a protocol of the Internet Engineering Task Force put in place to set standards in the dynamic registration of clients with authorisation servers. 
+We will go into the specifics of how it works in the context of Tyk, but if you are interested in reading the full RFC, go to: https://tools.ietf.org/html/rfc7591
+
+# Why should I use it? 
+
+DCR is a way for you to integrate your developer portal with an external identity provider such as Keycloak, Gluu, Auth0, Okta etc... 
+The portal developer won't notice a difference. However when they create the app via Tyk Developer portal, Tyk will dynamically register that client on your authorization server. This means that it is the Authorization Server who will issue issue the Client ID and Client Secret for the app.
+Some of our users leverage external Identity Providers because they provide a variety of features to support organisations in managing identity in one place across all their stack. 
+
+This feature is optional and you can still have a great level of security only using Tyk as your authorisation server. 
+
+# Enabling Dynamic Client Registration
+
+We provide guides for the following identity providers:
+
+- [Keycloak](/docs/tyk-developer-portal/keycloak-dcr/). Additional notes are found [here](https://github.com/keycloak/keycloak-documentation/blob/master/securing_apps/topics/client-registration.adoc).
+- [OKTA](/docs/tyk-developer-portal/okta-dcr/). Official docs are available [here](https://developer.okta.com/docs/reference/api/oauth-clients/).
+
+Support for Gluu has also been tested, you may find the docs [here](https://www.gluu.org/).
+
+In case your provider isn't on the list, use the "Other" provider option in the DCR settings. This mode would keep the interaction with your IDP as standard possible. Note that not all IDPs fully implement the standard.
+
+# Troubleshooting
+
+The DCR functionality abstracts most of the errors to the end user (in this case, the developer). In order to diagnose issues between Tyk and your IDP, please refer to the Tyk Dashboard logs.


tyk-docs/content/tyk-developer-portal/dynamic-client-registration.md
```diff
@@ -0,0 +1,105 @@
...
+weight: 1
+---
+
+## OAuth 2.0 Dynamic Client Registration Protocol (DCR)
+Available on version 3.2.0 onwards - currently not generally available but if you want beta access, get in touch! 
+
+# What is it? 
+
+DCR is a protocol of the Internet Engineering Task Force put in place to set standards in the dynamic registration of clients with authorisation servers. 
+We will go into the specifics of how it works in the context of Tyk, but if you are interested in reading the full RFC, go to: http://www.rfc-editor.org/info/rfc7591
```
asoorm: I would use this link instead. https://tools.ietf.org/html/rfc7591

tyk-docs/content/tyk-developer-portal/dynamic-client-registration.md
```diff
@@ -0,0 +1,105 @@
...
+
+# Why should I use it? 
+
+DCR is a way for you to integrate your developer portal with an external identity provider such as Keycloack, Gluu, Auth0, Okta etc...
```
asoorm: Keycloack -> Keycloak

tyk-docs/content/tyk-developer-portal/dynamic-client-registration.md
```diff
@@ -0,0 +1,105 @@
...
+
+# What is it? 
+
+DCR is a protocol of the Internet Engineering Task Force put in place to set standards in the dynamic registration of clients with authorisation servers. 
+We will go into the specifics of how it works in the context of Tyk, but if you are interested in reading the full RFC, go to: http://www.rfc-editor.org/info/rfc7591
+
+# Why should I use it? 
+
+DCR is a way for you to integrate your developer portal with an external identity provider such as Keycloack, Gluu, Auth0, Okta etc... 
+The user won't see a difference, but when they register a client on the portal, Tyk will proxy that request to your identity provider(IdP) and you will be able to manage their access crendentials either from the IdP or from Tyk.
```
asoorm: The portal developer won't notice a difference. However when they create the app via Tyk Developer portal, Tyk will dynamically register that client with your authorization server. This means that it is the Authorization Server who will issue issue the Client ID and Client Secret for the app.

tyk-docs/content/tyk-developer-portal/dynamic-client-registration.md
```diff
@@ -0,0 +1,105 @@
...
+We will go into the specifics of how it works in the context of Tyk, but if you are interested in reading the full RFC, go to: http://www.rfc-editor.org/info/rfc7591
+
+# Why should I use it? 
+
+DCR is a way for you to integrate your developer portal with an external identity provider such as Keycloack, Gluu, Auth0, Okta etc... 
+The user won't see a difference, but when they register a client on the portal, Tyk will proxy that request to your identity provider(IdP) and you will be able to manage their access crendentials either from the IdP or from Tyk. 
+
+Some of our users leverage external Identity Providers because they provide a variety of features to support organisations in managing identity in one place across all their stack. 
+
+This feature is optional and you can still have a great level of security only using tyk as your identity provider.
```
asoorm: Swap `identity provider` for `authorization server`. Tyk is not an Identity Provider and has no concept of identities. An Identity Provider is a special type of OAuth2 authorization server.

tyk-docs/content/tyk-developer-portal/dynamic-client-registration.md
```diff
@@ -0,0 +1,105 @@
...
+
+# I have an identity provider, how can connect it? 
+
+Tyk making the intermediation, we recommend that you also read the documentation of your identity provider as there will be some steps to follow on your IDP and the flow below assumes that you have completed the steps on your IdP.
+
+Gluu: https://www.gluu.org/how-to-dynamically-register-openid-connect-client/
+Okta: https://developer.okta.com/docs/reference/api/oauth-clients/
+Auth0: https://auth0.com/docs/applications/dynamic-client-registration
+Keycloack: https://www.keycloak.org/docs/latest/securing_apps/#openid-connect-dynamic-client-registration
+Other options include Azure, Amazon, Ping...
```
asoorm: Spelling: Keycloak
I would put this list in Alphabetical Order, with Other being last?

tyk-docs/content/tyk-developer-portal/dynamic-client-registration.md
```diff
@@ -0,0 +1,105 @@
...
+Tyk making the intermediation, we recommend that you also read the documentation of your identity provider as there will be some steps to follow on your IDP and the flow below assumes that you have completed the steps on your IdP.
+
+Gluu: https://www.gluu.org/how-to-dynamically-register-openid-connect-client/
+Okta: https://developer.okta.com/docs/reference/api/oauth-clients/
+Auth0: https://auth0.com/docs/applications/dynamic-client-registration
+Keycloack: https://www.keycloak.org/docs/latest/securing_apps/#openid-connect-dynamic-client-registration
+Other options include Azure, Amazon, Ping... 
+
+
+## Step 1: Create an API with JWT Authentification mode
```
asoorm: Authentification -> Authentication

tyk-docs/content/tyk-developer-portal/dynamic-client-registration.md
```diff
@@ -0,0 +1,105 @@
...
+If you're not sure how to do that, refer to : [https://tyk.io/docs/getting-started/tutorials/create-api/]
+In the "public key" section, make sure you enter your JWKS url. This should come from your identity provider.
```
asoorm: Screenshot 2021-02-23 at 20 24 10

In the OpenID Connect Standard - It's called the `jwks_uri`. Not `jwks_url`.

This should come from your identity provider. You would typically find it in the `.well-known/openid-configuration` endpoint.

tyk-docs/content/tyk-developer-portal/dynamic-client-registration.md
```diff
@@ -0,0 +1,105 @@
...
+Recommended
+* ``Authorization code`` Defined in https://tools.ietf.org/html/rfc6749#section-1.3.1
+* ``Client credentials`` Defined in https://tools.ietf.org/html/rfc6749#section-4.4
+* ``Refresh token`` Defined in https://tools.ietf.org/html/rfc6749#section-1.5
+* ``JWT Bearer token`` Defined in https://tools.ietf.org/html/rfc7523 Value is ``urn:ietf:params:oauth:grant-type:jwt-bearer``
+* ``SAML 2.0 Bearer Assertion`` Defined in https://tools.ietf.org/html/rfc7522 Value is ``urn:ietf:params:oauth:grant-type:saml2-bearer``
+
+Legacy
+* ``Implicit``Implicit grant type defined https://tools.ietf.org/html/rfc6749#section-1.3.2
+* ``Password`` Your IdP (resource owner) password credentials as a grant type https://tools.ietf.org/html/rfc6749#section-1.3.3
```
asoorm: Have each of these mechanisms been tested? 
It would be simpler to enable `Authorization Code`, `Client Credentials` and `Refresh Token` for MVP.

e.g. Why enable legacy Implicit & Password grants? Do we have a use-case for SAML 2 bearer assertion? How would that work? Do I still get a client_id and client_secret in the portal?

tyk-docs/content/tyk-developer-portal/dynamic-client-registration.md
```diff
@@ -0,0 +1,105 @@
...
+
+
+Your IdP may allow other values such as the device_code grant for application not using a browser or with constrained input. 
+
+3. Select a ``token endpoint auth method``
+
+Specify which authentication method should be used for the token endpoint. 
+* "None" if the client is public and does not have a 'secret'
+* "Client secret post" if the client uses the HTTP POST parameters
+* "Client secret basic" if the cleint uses HTTP basic
```
asoorm: I feel like this is wrong. Here - the person configuring DCR - should be specifying which token endpoint auth methods they will permit. e.g. Do I want to allow public clients? Should I allow auth requests using client secret post or client secret basic?

The portal developer would then when creating their client would be restricted in terms of what kind of client they can create, and how they intend to authenticate with the API.

tyk-docs/content/tyk-developer-portal/dynamic-client-registration.md
```diff
@@ -0,0 +1,105 @@
...
+* "None" if the client is public and does not have a 'secret'
+* "Client secret post" if the client uses the HTTP POST parameters
+* "Client secret basic" if the cleint uses HTTP basic 
+
+4. Select one or more ``Response type``
+
+``code`` used for the Authorization Code grant OAuth 2.0 flow. 
+``token`` is used for implicit grant. 
+
+There are more options for complex use cases, though you only need to pick Code or Token.
```
asoorm: Can this be blank for client_credentials?

tyk-docs/content/tyk-developer-portal/dynamic-client-registration.md
```diff
@@ -0,0 +1,105 @@
...
+
+5. Enter an ``Identity Provider host`` URL
+
+This is the URL of your IDP. For example https://gluu.do.myproject
```
asoorm: Why do I need this?

tyk-docs/content/tyk-developer-portal/dynamic-client-registration.md
```diff
@@ -0,0 +1,105 @@
...
+
+6. Enter a ``Client registration endpoint`` URL 
+
+Usually an extension of the Identity provider host URL
```
asoorm: You will find this in the `.well-known/openid-configuration` under `registration_endpoint`.

It might be easier to just request the user provide their `.well-known/openid-configuration` document, and we introspect it to mine the fields we need?

tyk-docs/content/tyk-developer-portal/dynamic-client-registration.md
```diff
@@ -0,0 +1,105 @@
...
+
+7. Initial ``registration access token`` (optional)
```
asoorm: We could call it a `registration access token`. What is `Initial`?

asoorm (changes requested):
Good start - a few comments & extra references: 

OpenID Connect Dynamic Client Registration 1.0 - for OpenID Connect relying parties. This spec should be used for almost all cases when we need to register apps with an OpenID Connect Identity Provider.

OAuth 2.0 Dynamic Client Registration Protocol - RFC 7591 - for simple OAuth 2.0 clients that don't need OpenID Connect features. e.g. mainly clients using the client credentials grant, but still need to be onboarded into an authorization server.

OAuth 2.0 Dynamic Client Registration Management Protocol - RFC 7592 - describes how to read, update or delete a client registration.

tyk-docs/content/tyk-developer-portal/dynamic-client-registration.md
gregdelhon: @matiasinsaurralde what do you think?

tyk-docs/content/tyk-developer-portal/dynamic-client-registration.md
gregdelhon: Not sure I understand, why do you think this is wrong? Are you saying we shouldn't get the admin to set that and keep it to "none" as default?

tyk-docs/content/tyk-developer-portal/dynamic-client-registration.md
gregdelhon: Good point, @ConsM did we make that field compulsory? I think we did but need to revert that and let users enable no entry.

tyk-docs/content/tyk-developer-portal/dynamic-client-registration.md
gregdelhon: @matiasinsaurralde this one is probs for you

tyk-docs/content/tyk-developer-portal/dynamic-client-registration.md
gregdelhon: Good suggestion, I am adding this to the backlog for when we move it to raava https://tyktech.atlassian.net/browse/TT-1696

tyk-docs/content/tyk-developer-portal/dynamic-client-registration.md
asoorm: The portal manager's role here is to decide or enforce what kind of clients & what auth mechanisms they wish to be enabled for the portal developer to choose from. So this should be a multi-select.

The portal developer knows whether they wish to create a public client or a confidential client.

---

So from the portal developer perspective, if there are multiple items checked - the portal developer should be able to create a public or confidential client - right from the developer portal. But if the `client secret post` or `client secret basic` are checked - then the portal developer should only be allowed to create a confidential client.

tyk-docs/content/tyk-developer-portal/dynamic-client-registration.md
```diff
@@ -0,0 +1,105 @@
...
+menu:
+  main:
+    parent: "Tyk Developer Portal"
+weight: 1
+---
+
+## OAuth 2.0 Dynamic Client Registration Protocol (DCR)
+Available on version 3.2.0 onwards - currently not generally available but if you want beta access, get in touch! 
+
+# What is it?
```
asoorm: Titles become anchor links

### What is it?

vs

### Dynamic Client Registration overview

tyk-docs/content/tyk-developer-portal/dynamic-client-registration.md
```diff
@@ -0,0 +1,105 @@
...
+
+## OAuth 2.0 Dynamic Client Registration Protocol (DCR)
+Available on version 3.2.0 onwards - currently not generally available but if you want beta access, get in touch! 
+
+# What is it? 
+
+DCR is a protocol of the Internet Engineering Task Force put in place to set standards in the dynamic registration of clients with authorisation servers. 
+We will go into the specifics of how it works in the context of Tyk, but if you are interested in reading the full RFC, go to: http://www.rfc-editor.org/info/rfc7591
+
+# Why should I use it?
```
asoorm, line 18: can delete this title and leave it as a couple of paragraphs in the overview section.

tyk-docs/content/tyk-developer-portal/dynamic-client-registration.md
```diff
@@ -0,0 +1,105 @@
...
+# Why should I use it? 
+
+DCR is a way for you to integrate your developer portal with an external identity provider such as Keycloack, Gluu, Auth0, Okta etc... 
+The user won't see a difference, but when they register a client on the portal, Tyk will proxy that request to your identity provider(IdP) and you will be able to manage their access crendentials either from the IdP or from Tyk. 
+
+Some of our users leverage external Identity Providers because they provide a variety of features to support organisations in managing identity in one place across all their stack. 
+
+This feature is optional and you can still have a great level of security only using tyk as your identity provider. 
+
+# I have an identity provider, how can connect it?
```
asoorm: ### I have an identity provider, how can connect it? 

vs

### Enabling Dynamic Client Registration

tyk-docs/content/tyk-developer-portal/dynamic-client-registration.md
matiasinsaurralde: I don't think we can select multiple options, at least on the IDPs that we've tried, you can only pick one, so you select either `client_secret_post` or `client_secret_basic`, etc.

marksou:
@matiasinsaurralde is happy so I'm merging this as 3.2 being released

PR #1568, opened 2021-03-02
Update supported Redis and MongoDB versions

n2linux:
Added the latest version of Redis and MongoDB that we support


marksou (approved):
LGTM

PR #1583, opened 2021-03-12
Open policy Agent presentation

lghiur:
Addresses https://tyktech.atlassian.net/browse/TT-1565


tyk-docs/content/getting-started/key-concepts/open-policy-agent.md
```diff
@@ -0,0 +1,148 @@
+---
+title: Open Policy Agent
+menu:
+  main:
+    parent: "Key Concepts"
+weight: 70
+---
+
+Tyk now supports the [Open Policy Agent](https://www.openpolicyagent.org/) (OPA) standard.
```
buger: I think it is misleading. 
People will think that we support it for the gateway. 

I do not think it is ready to be included into concepts section, and rather should be part of the "Tyk Stack" -> "Tyk Dashboard" section. Maybe even under RBAC section.

buger: Also lets clearly say that it is ONLY for the Tyk dashboard.

lghiur: @buger  if I mention that the APIs are part of Tyk Dashboard API would it be ok?

buger: Still I do not think that it should sit in the concepts at all. The Concept is RBAC, open-policy is just one of the implementations. Also I would move Opa configuration section after basic explanation of what Opa integration is, e.g. before `### Language intro` section. It will also fix this "lonely" one sentence intro at the start of doc. 

First paragraph should be really clear about what it is, and start with smth like:

> Tyk Dashboard permission system can be extended by writing custom rules using [Open Policy Agent](link) (in future OPA). Rule engine works on top of Dashboard API, which means you can control not only access rules, but also behaviour of all APIs (except public developer portal). 
>
> To give you some inspiration here are some ideas of the rules you can implement now:
>
> * Enforce HTTP proxy option for all APIs which target URL does not point to the internal domain
> * Control access for individual fields. For example, do not allow change API "active" status (e.g. deploy), unless you > have a specific permission set (and make new permission be available to the UI/API). Custom permissions can be 
 creating using the [Additional Permissions API](/docs/tyk-dashboard-api/org/permissions/)
> * Have a user(or group) which has read access to one APIs and write to another
>
> ### Configuration
> By default Dashboard OPA engine is turned off, and you need to explicitly enable it via configuration file.
> You can control OPA functionality on global level via configuration file or per organisation level using [OPA API](/docs/tyk-dashboard-api/org/permissions/) and [UI](#using-open-policy-agent-(opa)-in-the-dashboard). 
>
> #### Available configuration options
> | Key                             | Type       | Description                                                                                              | Example                 |
> | -------------------------------------- | ---------- | -------------------------------------------------------------------------------------------------------- | ----------------------- |
> | security.open_policy.enabled           | boolean    | Toggle support for Open Policy Agent                                                                     | true                    |
> | security.open_policy.debug             | boolean    | Enable debugging mode, prints a lot of information to the console                                        | true                    |
> | security.open_policy.disable_api       | boolean    | Disables access to the OPA API, even for users with Admin role (you can still maintain rules via file)                                           | true
> | security.additional_permissions        | string map | Add custom user/user_group permissions. You can use them in your rules, and they will be displayed on UI | `{"key": "human name"}` |
> 
> ### Language intro

buger: Also lets make a consitent use of OPA abbreviation (should be upper case everywhere).

lghiur: Agreed

buger: Also I do not see any reference to the fact that current dashboard rules specified in `schemas/dashboard.rego` file. 
And as suggested in another comment, I think we should create a separate page with annotated source code embeddedd to the docs (just put it inside ``` and hugo will do the rest). And probably we should have a link to it right from the first paragraph. Smth like:

When OPA rule engine is enabled, Tyk Dashboard use rules specified in [schemas/dashboard.rego][link] as reference RBAC implementation. On-premise users can directly modify this file to control dashboard behaviour.

tyk-docs/content/getting-started/key-concepts/open-policy-agent.md
```diff
@@ -0,0 +1,148 @@
...
+
+To give you some inspiration here are some ideas of the rules you can implement now:
+
+* Enforce HTTP proxy option for all APIs which target URL does not point to the internal domain
+* Control access for individual fields. For example, do not allow change API "active" status (e.g. deploy), unless you have a specific permission set (and make new permission be available to the UI/API). Custom permissions can be creating using the [Additional Permissions API](/docs/tyk-dashboard-api/org/permissions/)
+* Have a user(or group) which has read access to one APIs and write to another
+OpenPolicy rule engine put on top of Dashboard API, which means you can control the behavior of all APIs (except public developer portal)
+
+### Language intro
+OPA is purpose built for reasoning about information represented in structured documents. The data that your service and its users publish can be inspected and transformed using OPA’s native query language - Rego.
```
buger: Frankly it sounds a bit vague. I think official OPA docs do better job, and we can just copy paste it from their website:

> The Open Policy Agent (OPA, pronounced “oh-pa”) is an open source, general-purpose policy engine that unifies policy enforcement across the stack. OPA provides a high-level declarative language (Rego) that lets you specify policy as code and simple APIs to offload policy decision-making from your software.

tyk-docs/content/getting-started/key-concepts/open-policy-agent.md
```diff
@@ -0,0 +1,148 @@
...
+
+### What is Rego?
+Rego was inspired by Datalog, which is a well understood, decades old query language. Rego extends Datalog to support structured document models such as JSON.
```
buger: Lets add note from official docs on the top (it changed a lot since I worked with it last time, in a good way). 

> OPA policies are expressed in a high-level declarative language called Rego. Rego (pronounced “ray-go”) is purpose-built for expressing policies over complex hierarchical data structures. For detailed information on Rego see the [Policy Language](https://www.openpolicyagent.org/docs/latest/policy-language) documentation.

tyk-docs/content/tyk-stack/tyk-manager/open-policy-agent.md
```diff
@@ -0,0 +1,161 @@
...
+| security.additional_permissions        | string map | Add custom user/user_group permissions. You can use them in your rules, and they will be displayed on UI | `{"key": "human name"}` |
+
+
+With Opa turned on, the majority of the security rules will be dynamically evaluated based on these rules.
+
+Additionally, users can modify OPA rules, and define their own, through the [OPA API](/docs/tyk-dashboard-api/org/permissions/), or, On-premises users can access and modify the OPA rules from the file system, in [`schemas/dashboard.rego`](/docs/tyk-dashboard/opa-rules).
+Moreover, using these rules you can also modify request content 🚀
+Suggestion is to use those modifications in dev environments also do not forget to create a backup of the rego rules (ideally to store it in git)
+
+Enable flag is used on organisation level in case that it's not set or it's false we are aming settings from `tyk_analytics.conf`.
```
maciejwojciechowski, line 37: should we add some info here that `enable` flag on Organization level can be changed only by SuperAdmin over Dashboard Admin API ?

marksou:
After a conversation with @lghiur, we are closing this PR and creating a new one #1881

PR #1631, opened 2021-04-15
Clarify /oauth/authorize-client's api_id value for on-premise installations

jhass:
This is mentioned in an info box on https://tyk.io/docs/basic-config-and-security/security/authentication-authorization/oauth2-0/auth-code-grant/ but was otherwise extremely hard to discover from just looking at https://tyk.io/docs/basic-config-and-security/security/authentication-authorization/oauth-2-0/#authorisation-token-flow-eg-server-side-web-apps or these API docs.


marksou (approved):

marksou:
@jhass Thanks for contributing.

joshblakeley:
@marksou this also needs consultation before merge - can we revert please.

PR #1633, opened 2021-04-19
add a note about graphql api to dashboard keys and policies pages

devsergiy:
image

### DIFF
--- a/tyk-docs/content/tyk-apis/tyk-dashboard-api/api-keys.md
+++ b/tyk-docs/content/tyk-apis/tyk-dashboard-api/api-keys.md
@@ -330,7 +330,7 @@ authorization:7a7b140f-2480-4d5a-4e78-24049e3ba7f8
 
 ### Graphql API
 
-At the moment dashboard is using graphql api to operate keys
+Presently, the Tyk Dashboard uses the GraphQL API for keys.
 
 | **Method** | **URL**  | **Description**             |
 | ---------- | ------------- | --------------------------- |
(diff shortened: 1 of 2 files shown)


marksou (approved):

joshblakeley:
This read a bit oddly to me. Also the text is very short and uninformative. Can we consider enhancing before merge please - consider the view of a docs user seeing this with no context.

devsergiy:
@joshblakeley the point here was just to mention that we are using graphql API internally, but Leo didn't want to emphasize it

joshblakeley:
Thanks @spetrunin. I think it would be more useful to take an opinionated stance on this and communicate it for the end user such as:
```
In addition to the complete REST API for the Tyk Dashboard there is an experimental GraphQL API which is used by our frontend application for a smaller subset of the Tyk Dashboard capability. 

If you wish to experiment with this GraphQL API it can be interacted with on the routes listed below. We currently do not recommend using this GraphQL API over the REST API for CI/CD and automated tasks leveraging the Tyk Dashboard.
```

Does that make sense?

devsergiy:
@joshblakeley where do u want to put this?

@buger could you consult on this?

buger:
If you want to put it like this, ideally it needs be complemented with a proper GraphQL documentation, accessible from docs website. Also we need to check quality of this docs as well.

Additionally it probably should be a separate page with GraphQL API, living on the same level as Dashboard API link (no need to additional statements in keys and policies pages). Where we actually can put what Josh wrote, that it is in experimental form and REST is recommended.

devsergiy:
@joshblakeley I have created a separate task for creating better documentation for the graphql API it will be possible to do after having docs for the dashboard API https://tyktech.atlassian.net/browse/TT-2402

For now, we should go with such notes

@marksou could u merge this? or should I do this?

marksou:
@spetrunin I'll merge it now.

PR #1657, opened 2021-05-07
Simplify TTL expiration

buger:
We do not need multi-org capping as default option.


buger:
I will merge it for now, and @joshblakeley can check it after. We had multiple customers asking about it recently.

PR #1663, opened 2021-05-07
Update tyk-gateway-configuration-options.md

rewsmith:
Updated bind_to_slugs description since this no longer needs to be true for hybrid gateways

https://tyktech.atlassian.net/browse/TT-2445


marksou (approved):

buger:
🤔

PR #1687, opened 2021-05-31
Add kubernetes pages

sedkis:


marksou:
@letzya How is this working with what we are doing?

letzya:
Let me check it when I'm back

PR #1706, opened 2021-07-06
TT-2693 added example of grpc data streaming

sredxny:
added example for bidirectional data streaming in grpc, with service exposed in https but grpc server running in h2. This corresponds to the ticket https://tyktech.atlassian.net/browse/TT-2693

We didn't change any config, so it will work with what we already have documented for gRPC, hence just added an advanced example of gRPC


sredxny:
Any gramatical/semantic correction is appreciated!

marksou (approved):
👍

gregdelhon (commented):
Made some minor grammar improvements, otherwise this is great and ready to go!

gregdelhon:
@matiasinsaurralde have a quick look, and then let's merge :)

PR #1711, opened 2021-07-15
fix redis duplication

letzya:


marksou (approved):
As per @letzya request.

PR #1712, opened 2021-07-15
Fix RH & Ubuntu OSS links

marksou:


marksou:
This has been fixed in other PRs, so closing

PR #1746, opened 2021-09-03
[TD-585] Remove helm chart version

DeployInBinary:
This PR is removing the need to specify helm chart version when installing our helm charts. If no version is specified then the latest one will be used.


marksou (approved):
LGTM

PR #1749, opened 2021-09-06
Add mention of MongoDB URL

marksou:


rewsmith (changes requested):
I think this note would be better included within the existing MongoDB paragraph 

Follow notes from the installation output to get connection details and password. The DNS name of your MongoDB as set with Bitnami is tyk-mongo-mongodb.tyk.svc.cluster.local You can update them in your local values.yaml file under mongo.mongoURL Alternatively, you can use --set flag to set it in Tyk installation.

Suggest amending to ...

Follow notes from the installation output to get connection details and password. The DNS name of your MongoDB as set with Bitnami is `tyk-mongo-mongodb.tyk.svc.cluster.local` and you also need to set the `authSource` param to `admin`.  The full mongoURL should be similar to `mongoURL: mongodb://root:pass@tyk-mongo-mongodb.tyk.svc.cluster.local:27017/tyk_analytics?authSource=admin`. You can update them in your local values.yaml file under mongo.mongoURL Alternatively, you can use --set flag to set it in Tyk installation.

rewsmith:
The block of text you originally added is now redundant - so should be removed

When using the MongoDB bitnami chart, you need to set your MongoDB URL in your values.yaml file with authSource set. For example:

`mongoURL: mongodb://root:pass@tyk-mongo-mongodb.tyk.svc.cluster.local:27017/tyk_analytics?authSource=admin`

rewsmith (approved):

PR #1759, opened 2021-09-21
Add instructions on how to connect to atlasdb

sedkis:
Add instructions for example config files on conencting to AtlasDB because it is actually not very straightforward to the SRV setup.


marksou (approved):
LGTM

PR #1765, opened 2021-09-27
Curity DCR integration in Dev Portal

iggbom:
Adds documentation on how to configure the Curity Identity Server and Tyk to achieve DCR integration with the Tyk Dev Portal. Uses either JWT or Split Token to protect the API.


gregdelhon (approved):
LGTM 🎉 
Thanks again @iggbom

PR #1768, opened 2021-09-30
Removed page and added content to top level page

marksou:
Moved content to the top-level page and did a bit of editing.


craigtyk (approved):
LGTM, thanks

PR #1776, opened 2021-10-06
[TT-3523] Add github action to update swagger.yml from gateway

jeffy-mathew:
Currently [gateway-swagger.yml](https://github.com/TykTechnologies/tyk-docs/blob/master/tyk-docs/static/others/gateway-swagger.yaml) in tyk-docs and [swagger.yml](https://github.com/TykTechnologies/tyk/blob/master/swagger.yml) in gateway are disconnected. tyk-docs `gateway-swagger.yml` needs to be updated manually each time when gateway swagger is updated. 

This PR intents to add a job in `config-gen` action so that the contents of gateway `swagger.yml` is added to tyk-docs `gateway-swagger.yml` when the `Config Generator` workflow is dispatched.

Documentation and swagger will be updated through two separate commits


.github/workflows/config_gen.yaml
```diff
@@ -22,22 +22,48 @@ jobs:
...
       run: |
+          cd ./tyk-docs
           git config --global user.name 'Leonid Bugaev'
           git config --global user.email 'leonsbox@gmail.com'
           git add -A
-          git commit -m "Updated configuration documention $REPO:$BRANCH"
+          git diff-index --quiet HEAD || git commit -m "Updated configuration documention $REPO:$BRANCH"
           git push
+
+    - name: Checkout gateway
```
buger, line 47: Once nuance, it should run only if $REPO is gateway. 
GitHub actions support conditionals, so should not be a problem.

.github/workflows/config_gen.yaml
```diff
@@ -22,22 +22,48 @@ jobs:
...
+      uses: actions/checkout@master
+      with:
+        repository: TykTechnologies/tyk
+        path: ./tyk
+        token: ${{ secrets.REPO_TOKEN }}
+        ref: ${{ github.event.inputs.repoBranch }}     
+
+    - name: Copy Swagger
+      run: |
+        cat ./tyk/swagger.yml > ./tyk-docs/tyk-docs/static/others/gateway-swagger.yml
```
buger, line 59: can be replaced with `cp`

.github/workflows/config_gen.yaml
jeffy-mathew: Added the condition to steps used for swagger generation

PR #1787, opened 2021-10-15
Add version cavaet to docs

sedkis:
We released the docs on go native plugins but they're not released until v4.0.0, just adding that cavaet.


tbuchaillot (approved):

sedkis:
@marksou  G2G

PR #1799, opened 2021-10-27
Clarifying Language for Custom Domains

nerdydread:
Added clarifying language to custom domains and edge gateways.


marksou:
@nerdydread Who is best to review this from a technical viewpoint?

excieve (commented):
Thanks @nerdydread! This is the case for both the Edges and the Control Planes and between them too. In other words, a custom domain must be unique for any deployment.

craigtyk (approved):
lgtm

PR #1800, opened 2021-10-28
[TT-1501] Add Portal set up links

marksou:


rewsmith (changes requested):
Hi Mark.  I'm guessing it's not possible to link to the Self-Managed Dev Portal Tutorial tab??  If not possible then I think we should explicitly state in the link to follow the Self-Managed instructions.

marksou:
Hi @rewsmith you are correct. I'll add something about the self-managed tab.

rewsmith (approved):

PR #1802, opened 2021-11-03
Add template archetypes for concepts, tasks, reference

marksou:


dcs3spp (commented):
@letzya This looks to be technical writing style guidelines. Instead of having it in the tyk-docs repo I think this content would be more suitable for use within the Docs Bible. What do you think?

letzya:
@dcs3spp closing for now. Just know this is here in case you want to start adding this. We could add the release-notes template as an archetype.

PR #1827, opened 2022-01-11
[TT-3919] Drop buster (oldstable) and stretch (oldoldstable) from uploads

alephnull:
Not longer supported by the latest releng.


marksou (approved):
👍

PR #1829, opened 2022-01-14
Update config options of gateway

sredxny:
Update the configuration options of gateway, I noticed that `key_space_sync_interval` is not available in the documentation currently on-line


sredxny:
closed as is documented in master but not released to the website

PR #1837, opened 2022-01-28
[TT-4363] Document EL7 plugin build

alephnull:
From 4.0 on, RHEL 7 and Xenial users will have to use a specially tagged plugin compiler.


marksou (approved):
LGTM

marksou:
Closing after confirming with @alephnull

PR #1890, opened 2022-05-03
[TT-4452]

AMM55:
Documentation update to notify that TLS 1.3 does not support cipher


marksou (changes requested):
Hi Ati, can you change the Note to our usual format, for example:

{{< note success >}}

**Note**  

 

The Tyk Pump is not currently configurable in our Tyk Cloud solution.

{{< /note >}}

marksou (approved):
Looks good now!

PR #1893, opened 2022-05-09
Moesif pump updates

matthewtanner91:


tbuchaillot (changes requested):
Hi @matthewtanner91! Can you explain the changes of the config files for `.env` ones? I think if we want to include the env vars here, adding a "Setting X in your config file or Y env vars" should be better than just removing the config options from our docs.

sedkis:
Updated the doc to include both JSON and ENV config methods :) @tbuchaillot  @matthewtanner91

tbuchaillot (approved):

marksou (approved):

PR #1903, opened 2022-05-26
Adding example for vault integration and tyk-gateway env variables. 

JRWu:
Adding example of referencing HashiCorp Vault secrets through Tyk-Gateway and environment variables. 

Additionally removed dangling reference to old istio documentation causing doc build to fail.


marksou (approved):
LGTM

marksou:
@JRWu is this relevant for all currently supported versions?

JRWu:
@marksou  It's tested with 4.0 but I don't see any changed functionality in the source code for kvStore which is the vault implementation since January of 2020. I can test it out with older versions just to confirm and get back to you.

Additionally the env variables for vault are documented for 3 LTS, 3.1, and 3.2.

PR #1909, opened 2022-06-01
Update Helm Installation guide for tyk-pro

komalsukhani:
Added information about supported mongo versions.


buraksekili:
Should we change [this `tab` section](https://github.com/TykTechnologies/tyk-docs/blob/komalsukhani-patch-1/tyk-docs/content/tyk-on-prem/installation/kubernetes/helm-chart.md?plain=1#L109-L114) regarding SQL installation @komalsukhani? I realized it while looking at the [Installing the Data Sources](https://tyk.io/docs/tyk-self-managed/tyk-helm-chart/#installing-the-data-stores) section that shows the missing information for SQL installation.

komalsukhani:
@buraksekili  Handled in #1935

marksou (approved):

PR #1947, opened 2022-08-01
fix(request-plugins):  middleware execution order link

armujahid:
request lifecycle link at https://tyk.io/docs/plugins/request-plugins/ is currently pointing to https://tyk.io/docs/getting-started/key-concepts/middleware-execution-order/ which doesn't exist. It should point to https://tyk.io/docs/concepts/middleware-execution-order/


pagemag:
Hi @armujahid , Thank you for reporting this issue we will review this and do the necessary fix.

marksou (approved):

PR #1953, opened 2022-08-12
Update gateway-config.md

rewsmith:
Corrected the format of TYK_GW_SECURITY_CERTIFICATES_UPSTREAM env var

Slack thread: https://tyktech.slack.com/archives/C7E4B7V7Z/p1660301729147979


tbuchaillot:
@rewsmith those configs are generated from the code - it would be better to raise a Tyk Gateway PR with the changes in the config comments

rewsmith:
> @rewsmith those configs are generated from the code - it would be better to raise a Tyk Gateway PR with the changes in the config comments

Understood @tbuchaillot

rewsmith:
As per comment from @tbuchaillot , this should be a Gateway PR

PR #1974, opened 2022-08-26
documented new behavior while loading the plugins

sredxny:
Documented new behavior of the gateway when it loads the go plugins. This is regarding https://tyktech.atlassian.net/browse/TT-4354


buger:
Lets put your changes under "Managing Upgrades for Tyk" section or similar (there should be better name).

marksou (approved):

PR #1975, opened 2022-08-27
added mdcb as a valid value for config generator

sredxny:
Added the word MDCB as an allowed value for the config generator, so people don't get confused about using sink, tyk-sink or mdcb.


marksou (approved):
👍

letzya (approved):

PR #2006, opened 2022-09-12
added diagram that displays how goplugins are loaded since v4.1.0

sredxny:
Added diagram to show the behavior of the goplugin loader


marksou (approved):
LGTM

buger (approved):

PR #2022, opened 2022-10-11
more search results

buger:


marksou:
@buger Do we need to upgrade to V3 of docsearch? See https://docsearch.algolia.com/docs/migrating-from-v2

marksou (approved):

PR #2035, opened 2022-10-20
[TT-6708] Simplify OAS Docs

marksou:
As per the ticket


remotelycoding (approved):
Covers what I raised in the ticket, thank you.

PR #2062, opened 2022-11-15
Add Canonical tag

buger:
We do not have duplicate content, accessed though the different URLs, we do use redirects instead. 
Canonical tag will do nothing in this case, but it will help make automatic scanning tools pass all checks.


marksou (approved):
As discussed with Leo

madebyguie (approved):

PR #2063, opened 2022-11-15
Fix javascript missing source map warning

buger:


marksou (approved):
As discussed with Leo

PR #2077, opened 2022-11-23
[DX-154] helm 3 vrsion

AMM55:
Declare the Helm 3+ version in the docs


carolwkc:
Hi @AMM55 , thanks for update. 
On self managed / cloud page, Helm version support should be a separate point on its own.
Suggest to update to "Tyk Helm Chart supports the Helm 3+ version."

carolwkc (approved):

PR #2087, opened 2022-11-24
Add text below heading for search purposes.

marksou:
We needed to add text below the heading to enable the search crawler to pick up the heading.


zalbiraw (approved):
LGTM

PR #2106, opened 2022-12-06
Updating dashboard's OAS

letzya:
I did a copy/paste of https://github.com/TykTechnologies/tyk-analytics/blob/release-4.3/swagger.yml (please note, it's from 4.3 branch and not master, just in case)

The gateway OAS is actually up to date 

@furkansenharputlu @lghiur there are lots of changes, kindly advise on this PR


letzya:
@buger closing since there's zero file. Where is that file

PR #2111, opened 2022-12-06
[TT-1047] clarify tag_headers explanation

letzya:
Add link to full explanation


sedkis:
@letzya , not sure how to add the suggestion directly to the repo like you do, will put in in this command instead.

How about adding a sample use case, after the initial paragraph?


> One common use case for this would be to query aggregates for a certain HTTP Header.
Imagine you have a `tenant-id` or a `user-id` Header.  Well, after instructing Tyk to capture this via `tag_headers`, you can query to see count of all API requests made that included header IDs for each of your individual tenants or users.

letzya:
@sedkis thanks for th review. I am referring the readers to another page for further reading. This page is only about tag_header. https://tyk.io/docs/tyk-apis/tyk-gateway-api/api-definition-objects/custom-analytics/ - maybe we should add it there?

sedkis:
> @sedkis thanks for th review. I am referring the readers to another page for further reading. This page is only about tag_header. https://tyk.io/docs/tyk-apis/tyk-gateway-api/api-definition-objects/custom-analytics/ - maybe we should add it there?

Agreed - sounds good !

sedkis (approved):

PR #2125, opened 2022-12-12
Sync URL and File structure

buger:
Aim here is to sync URL structure with File structure. 
Added `./scripts/sync_files_with_paths.sh` which should automatically migrate majority of the files (which I did as part of this PR). There were a few manual obvious fixes. 

Additionally all URLs converted to use "ref" which also means that all relative urls now properly work with nightly and etc. Also we get validation of URL correctness on Hugo level. 

Added htmltest CI, which also can be run locally by running `bash ../script/htmltest.sh` inside tyk-docs folder.


letzya:
@buger we changed here upstream files (i.e. they are auto-generated in another place).

PR #2128, opened 2022-12-14
Error Response Codes Mapping

dle519:
Very rough look of all the information obtained so far currently.


tyk-docs/content/error-codes/error-codes.md
```diff
@@ -0,0 +1,229 @@
+---
+title: "Error Codes"
+date: 2022-12-13
+tags: [""]
+description: "Response codes that the gateway returns via an API"
+weight: 0
+draft: true
+---
+
+**Error Code 400:**
```
letzya:
```suggestion
## Error Code 400:
```

letzya (dismissed):
Please update to use headings (##) instead of bold font (**)

letzya:
@dle519 please assign this PR to a peer engineer for review

vverbani:
@dle519 not authorized to merge this PR unfortunately

tyk-docs/content/error-codes/error-codes.md
dle519: Updated!

tyk-docs/content/error-codes/error-codes.md
```diff
@@ -0,0 +1,229 @@
+---
+title: "Error Codes"
+date: 2022-12-13
```
JRWu: Is this date accurate? Perhaps it would be more informative to specify what version(s) of Tyk this document covers.

tyk-docs/content/error-codes/error-codes.md
```diff
@@ -0,0 +1,229 @@
...
+tags: [""]
+description: "Response codes that the gateway returns via an API"
```
JRWu: I believe our stance is to use "Tyk Gateway", or capitalize Gateway when referencing it

https://tyktech.atlassian.net/wiki/spaces/MAR/pages/1428848683/Copywriting+at+Tyk+Our+house+style

tyk-docs/content/error-codes/error-codes.md
```diff
@@ -0,0 +1,229 @@
...
+- ErrImportWithTykExtension
+- ErrPayloadWithoutTykExtension
+- Couldn’t decode OAS object
+- Batch request malformed
+- Batch request creation failed, request structure malformed
+- Error parsing form. Form malformed
+- Access to this API has been disallowed
+- Authorization field missing
+- Bearer token malformed
+- TBD
```
JRWu: What is TBD?

tyk-docs/content/error-codes/error-codes.md
```diff
@@ -0,0 +1,229 @@
...
+**Error Code 401:**
+
+- Authorization Field Missing
+- Wrong Password (Basic Auth)
+- Header missing (JS Middleware)
+- Not authorised (JS middleware)
+- oauthClientIdEmpty
+- oauthClientSecretEmpty
+- Request signature verification failed (auth keys: could be empty signature header, or validation failed)
+- TBD
```
JRWu: TBD Again, this acronym needs to be expanded upon or removed

tyk-docs/content/error-codes/error-codes.md
```diff
@@ -0,0 +1,229 @@
...
+- Spec not valid, skipped!
+- Body do not contain username (Basic Auth from body)
+- Body do not contain password
+- Request with empty authorization header (JWT session)
+- The provided request is empty (graphql playground)
+- Key cannot be used without a certificate
+- Key must be used with an existent certificate
+- Failed to remove the key
+- Request ID does not match that in policy! For Update operations these must match
+- ErrAPINotMigrated (oasEndpoint)
```
JRWu: New users aren't going to understand what oasEndpoint is, perhaps elaborate on this point further?
What is this error?

tyk-docs/content/error-codes/error-codes.md
```diff
@@ -0,0 +1,229 @@
...
+- Request APIID does not match that in Definition! For Update operations these must match
+- Must specify an apiID to update
+- Must specify an apiID to delete
+- Must specify an apiID to patch
+- Couldn’t decode instruction (policyUpdateHandler)
```
JRWu: What is policyUpdateHandler? This should be more informative to the end user; what instruction couldn't be decoded?

tyk-docs/content/error-codes/error-codes.md
```diff
@@ -0,0 +1,229 @@
...
+  - process request live with rate limit
+  - process request off thread with rate limit
+- This organisation access has been disabled, please contact your API administrator (orgSession.isInactive)
+- Depth limit exceeded (on an API level or Field Level)
+- Accessing to sub-version with base API listen path should require base API key (version checking)
+- Version Information not found
+- Client authorize request in with invalid redirect URI
+- Attempted administrative access with invalid or missing key! (checking correct security credentials of the tyk API)
+- Run Go-plugin auth failed (invalid token)
+- TBD
```
JRWu: TBD again

tyk-docs/content/error-codes/error-codes.md
```diff
@@ -0,0 +1,229 @@
...
+- This organisation rate limit has been exceeded, please contact your API administrator
+- This organization access has been disabled or quota/rate limit is exceeded, please contact your API administrator (processing requests off thread)
+- Invalid Token (Form Value Test for Value ExtractorHeader Source)
+- Tls: bad certificate (no valid cert IDs exist in Certificate.ControlAPI)
+- Client TLS certificate is required (Multiple APIs on same domain without certs)
+- Certificate with SHA256 $CLIENTCERT not allowed (Multiple APIs with Mutual TLS on same domain)
+- Unknown cert without domain
+- Unknown cert with custom domain
+- Key not authorised
+  - JWT → id is in the ‘sub’ claim while the ‘kid’ claim in the header is not empty
```
JRWu: Why is there tabbing here but not in the other error codes? Is there a reason why the formatting is different

JRWu (dismissed):
This is a good start at documenting the error codes. I think the user-added comments within parenthesis need to be elaborated on a bit further- think about it from the perspective of a new user. Are they going to understand what "policyUpdateHandler" means, or do they need more clarification? 

The tabbing is inconsistent- why is it only tabbed in some explanations but not other?

Perhaps try sorting the error messages per response code alphabetically- that way it's easier for users to read and consume. 

Additionally, and this might be a different piece of work- perhaps we should be adding potential resolutions to each error code, or descriptions on why they occur. End users who search for these error codes should be given a path to resolve the error- either through links to our documentation or a FAQ section illustrating the fix.

tyk-docs/content/error-codes/error-codes.md
dle519: I dont know what the exact acronym stands for but I read in the documentation that its convention for when the section is a WIP (work in progress).

vverbani: to be determined?

JRWu: If you've added all of the error codes & their respective messages TBD can be removed, it's not necessary to keep it here.

dle519: The information so far is all the error codes that I have found so far so there is still quite a lot of error codes that I haven't discovered yet. We are trying to publish what we have so far because clients have been asking for it so we thought a WIP might help them. Thats why the formatting of this looks very very rough at the moment since I am focusing on just getting the data out there.

oluwaseyeayinla:
@JRWu Good job. I like this approach showing relevant information very quickly. I think we should make some of the errors a bit more descriptive and readable. Things like testGRPCIgnore - no header, ErrImportWithTykExtension etc.

I also think we can expatiate on potential causes for very generic error messages e.g. Access to this API has been disallowed. I see you have done a couple with Key not authorised, OAuth and versioning. 

Access to this API has been disallowed
- Check if API definition is disabled
- Check if you are using the correct key/auth token
- Refer to some other error page or a doc link (if all fails)

The main aim would be to try not to repeat ourselves all over the page, since some errors have similarities in what to check or validate

tyk-docs/content/error-codes/error-codes.md
dle519: added an acronym in parenthesis for it

tyk-docs/content/error-codes/error-codes.md
dle519: changed the format to be more consistent

vverbani:
1. */error-codes/error-codes/ URL should be error-response-codes
- This gets rid of the duplicate error-codes plus it better outlines the page for SEO/user readability purposes

2. I don't like how this page is listed as WIP and TBD (to be determined?) - See Jia's comment about making WIP part of the PR title as opposed to the page title for future ref
we're in charge of this, clients don't need to know this portion imo. We can continue to work this on the backend and push content for this page after publishing the initial draft

- Instead, we can add a general statement something along the lines of ** as Tyk continues to scale, so will the error response codes. As such, we'll going to do our best to coordinate them on this page**

3. I don't like how the tables look. Some have descriptions while others don't. The part that mostly bothers me is that the same table is scattered. I think for a UI perspective it would look better if we did one of the following:

- List all under one table, but move all the response messages with their descriptions to the top while the ones without descriptions are all moved to the bottom
- Divide them into two tables, first with a description while second tables have no descriptions
Also why were some of these left blank. I realize they're self explanatory but could easily just write things like Your request is missing the 'api_id' key value pair

tyk-docs/content/error-response-codes.md
```diff
@@ -0,0 +1,121 @@
...
+description: "Response codes that the Tyk Gateway (v4.3.0) returns via an API"
+weight: 201
+menu:
+  main:
+    parent: Tyk Gateway
+---
+
+This table contains the error codes that could possibly show up when using the Gateway. Each error code has a text and a description of what could have caused the error. **As Tyk continues to grow, so will the error response codes. As such, we will do our best to coordinate them on this page.**
+
+| Code: | Text                                                                                    | Description                                                                                                                                                                                                                                                                                                                                                                                                      |
```
vverbani: Just Code, not Code:

tyk-docs/content/error-response-codes.md
```diff
@@ -0,0 +1,121 @@
...
+| 507   | Status Insufficient Storage                                                             | Attempted to update an API through a POST request but failed to due insufficient storage.                                                                                                                                                                                                                                                                                                                        |
+| x509  | Certificate signed by unknown authority                                                 | A client accessing Tyk with full ControlAPI access without a certificate.                                                                                                                                                                                                                                                                                                                                        |
```
vverbani: Not a response code so I would get rid of x509

vverbani (changes requested):
Just these 2 changes then it can be updated with branch / merged into branch 

@dle519 @letzya

tyk-docs/content/error-response-codes.md
letzya: Not sure I understand

tyk-docs/content/error-response-codes.md
```diff
@@ -0,0 +1,121 @@
...
+date: 2023-1-6
+tags: ["error codes", "error response codes"]
+description: "Response codes that the Tyk Gateway (v4.3.0) returns via an API"
+weight: 201
+menu:
+  main:
+    parent: Tyk Gateway
+---
+
+This table contains the error codes that could possibly show up when using the Gateway. Each error code has a text and a description of what could have caused the error. **As Tyk continues to grow, so will the error response codes. As such, we will do our best to coordinate them on this page.**
```
letzya: Please reword to a formal tone

letzya: maybe @chrisanderton  can advise

tyk-docs/content/error-response-codes.md
letzya:
```suggestion                                                                                                                                                                                                                                                                                                                               |
```

tyk-docs/content/error-response-codes.md
letzya:
```suggestion
| Code | Text                                                                                    | Description                                                                                                                                                                                                                                                                                                                                                                                                      |
```

tyk-docs/content/error-response-codes.md
chrisanderton: Should we call them "Error response _status_ codes" to be consistent with [HTTP language](https://httpwg.org/specs/rfc9110.html#overview.of.status.codes)?

We could probably borrow some of that for the intro too..? Could also split the table into client errors and server errors.

Probably shouldn't say "we'll do our best" as it doesn't inspire confidence. In the olden days, this kind of table would be generated directly from code, as the dictionary of codes and descriptions would be in a single file with constants and used consistently.. do we do that here?

<hr/>

Responses from the Gateway include a three-digit status code that describes the result of the request and the semantics of the response. The first digit of the status code defines the class of response. The error status codes used by the Gateway are defined in the table below.

- **1xx (Informational):** The request was received, continuing process
- **2xx (Successful)**: The request was successfully received, understood, and accepted
- **3xx (Redirection)**: Further action needs to be taken in order to complete the request
- [**4xx (Client Error)**](): The request contains bad syntax or cannot be fulfilled
- [**5xx (Server Error)**](): The server failed to fulfill an apparently valid request

dle519: To my knowledge, i dont think theres a table that can be generated directly from the code. These codes were manually found within the code so there could be still some error codes missing from this table.

vverbani: This isn't auto generated. It would be amazing if it was - definitely feature request Jira ticket worthy for this (David maybe you can create a ticket for this).

This PR was more of to help our customers during their environment set up/debugging until we get a more automatic version up and running. On the Support/Post-sales side we've had a lot of customers ask for this

vverbani: @letzya @dle519 @chrisanderton what are your thoughts? Can we move forward with the merge ?

dle519: looks good to merge for me!

letzya:
@dle519 please get peer review approval and change to "ready for review" if it's ready.

tyk-docs/content/error-response-codes.md
```diff
@@ -0,0 +1,126 @@
+---
+title: "Error Response Status Codes"
+date: 2023-1-6
+tags: ["error codes", "error status response codes"]
+description: "Error response codes that the Tyk Gateway (v4.3.0) returns via an API"
```
letzya:
```suggestion
description: "Error response codes that the Tyk Gateway returns via an API"
```

tyk-docs/content/error-response-codes.md
```diff
@@ -0,0 +1,126 @@
...
+weight: 201
+menu:
+  main:
+    parent: Tyk Gateway
+---
+
+Responses from the Gateway include a three-digit status code that describes the result of the request and the semantics of the response. The first digit of the status code defines the class of response. The error status codes used by the Gateway are defined in the table below.
```
letzya:
```suggestion
Tyk gateway responses include HTTP status codes that follow the [HTTP standard](https://datatracker.ietf.org/doc/html/rfc9110). They have three digits that describe the result of the request and the semantics of the response. The first digit defines the class of response as shown in the [list](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes) below:
```

tyk-docs/content/error-response-codes.md
```diff
@@ -0,0 +1,126 @@
...
+
+- 1xx (Informational): The request was received, continuing process
+- 2xx (Successful): The request was successfully received, understood, and accepted
+- 3xx (Redirection): Further action needs to be taken in order to complete the request
+- 4xx (Client Error): The request contains bad syntax or cannot be fulfilled
+- 5xx (Server Error): The server failed to fulfill an apparently valid request
+
```
letzya, line 20:
```suggestion

The error status codes (4xx and 5xx) used by Tyk Gateway are defined in the table below:
```

vverbani (approved):

PR #2129, opened 2022-12-15
Update ce-helm-chart.md

letzya:
add command to add redis to helm local directory


buraksekili (approved):
Also, it'd be better if we add the same instruction to [tyk-self-managed/tyk-helm-chart.md](https://github.com/TykTechnologies/tyk-docs/blob/master/tyk-docs/content/tyk-self-managed/tyk-helm-chart.md)

PR #2151, opened 2023-01-03
Fix compare button broken link

yurisasuke:
When the compare button is clicked it takes you to https://tyk.io/apim. instead of https://tyk.io/docs/apim/. I have fixed this issue so that you can be directed to correct Url.


letzya:
@Keithwachira closing this PR since it was resolved in a different way

PR #2154, opened 2023-01-04
Merging to release-4.3: fix broken links for all buttons (#2153)

buger:
fix broken links for all buttons (#2153)

When the compare button is clicked it takes you to https://tyk.io/apim.
instead of https://tyk.io/docs/apim/. I have fixed this issue so that
you can be directed to correct Url.FYI all button had been previously
broken this pull request will fix all buttons


letzya:
thanks @buger !!!

PR #2160, opened 2023-01-05
update self managed image.

yurisasuke:
The self managed image should show that we  support both mongo and sql. This pull request replaces the image that only had mongo to one with mongo and sql.


letzya (approved):
thanks

PR #2208, opened 2023-01-13
added section in our tyk pump on our tyk analytics fields

LLe27:
added details for each analytics data field for our schema sent from our tyk pump


tbuchaillot (approved):
LGTM

letzya:
Sorry for suggesting it only now but it looks like the same use case as our other configs which we autogen from the code? @zalbiraw what do you think?

tyk-docs/content/tyk-stack/tyk-pump/tyk-analytics-schema.md
```diff
@@ -0,0 +1,187 @@
+---
+title: "Tyk Analytics Schema"
```
letzya: Per discussion, change schemas to fields in the file name, title and everywhere else.

letzya (changes requested):
Per discussion, change schemas to fields

tyk-docs/content/tyk-stack/tyk-pump/tyk-analytics-schema.md
LLe27: Added the requested change @letzya. Please review thank you!

letzya (approved):

PR #2227, opened 2023-01-19
Fix custom_analytics page link in api-def-common.md

tbuchaillot:
## Description

Change custom_analytics link to its alias to fix CI problem https://github.com/TykTechnologies/tyk-docs/blob/master/tyk-docs/content/tyk-apis/tyk-gateway-api/api-definition-objects/custom-analytics.md

## Types of changes

- [ ] Fixing typo (please merge to production)
- [ ] Documenting a new feature (please merge to production)
- [ ] Documentation for future release (please do not merge to production)
- [ ] Something else (please add if needs merging to production or not)

## Checklist

- [ ] I have [reviewed the guidelines](../CONTRIBUTING.md) for contributing to this repository.
- [ ] I have [read the technical guidelines](../CONTRIBUTING-TECHNICAL-GUIDE.md) for contributing to this repository.
- [ ] Make sure you have started *your change* off *our latest `master`*.
- [ ] I have added a preview link.


tyk-docs/content/shared/api-def-common.md
```diff
@@ -31,7 +31,7 @@ Your Tyk Gateway can listen on multiple domains/subdomains through the use of re
...
+* `tag_headers`: This specifies a string array of HTTP headers values which turned to tags. For example if you include `X-Request-ID` header to `tag_headers`, for each incoming request it will include a `x-request-id-<header_value>` tag to request an analytic record. This functionality can be useful if you need analytics for request's headers without the body content (Enabling detailed logging is another option, but it records the full request and response objects and consume a lot more space). Check the page [custom Analytics Tags using HTTP Headers](/tyk-rest-api/api-definition-objects/custom-analytics/) for more important information.
```
letzya:
```suggestion
* `tag_headers`: This specifies a string array of HTTP headers values which turned into tags. For example, if you include the `X-Request-ID` header to `tag_headers`, for each incoming request it will include an `x-request-id-<header_value>` tag to request an analytic record. This functionality can be useful if you need analytics for request headers without the body content (Enabling detailed logging is another option, but it records the full request and response objects and consumes a lot more space). Check the page [custom Analytics Tags using HTTP Headers]({{< ref "tyk-apis/tyk-gateway-api/api-definition-objects)/custom-analytics" >}}) for more important information.
```

letzya: We need to use `ref`

letzya (changes requested):

tyk-docs/content/shared/api-def-common.md
tbuchaillot: @letzya now we got this error https://github.com/TykTechnologies/tyk-docs/actions/runs/3960292373/jobs/6784245686#step:4:9

I don't think that's the right way ?

letzya: Also, we should use the page path as the link, not the alias `tyk-apis/tyk-gateway-api/api-definition-objects)/custom-analytics` we only keep aliases because of google.

tbuchaillot: @letzya so we should use ref with the old link? `/docs/tyk-apis/tyk-gateway-api/api-definition-objects/custom-analytics/`? do we need to include the `.md` extension?

letzya: I updated my suggestion you can just commit it

letzya (approved):

PR #2260, opened 2023-01-26
Enable external URL check

buger:
Enable external URL check on CI


letzya:
Should we do it in a separate job and run it once a day?

letzya:
Per @buger - needs caching

dcs3spp:
Closing PR, a ticket has been created with Jira ID DX-422

PR #2550, opened 2023-05-16
Added error responses for size limit checks

andyo-tyk:
Added HTTP error codes and status messages from the size limit checking middleware here: https://github.com/TykTechnologies/tyk/blob/29d50289c48ba2fccacba5306dc8fb408be15e85/gateway/mw_request_size_limit.go#L35


dle519 (approved):

tyk-docs/content/error-response-codes.md
```diff
@@ -52,6 +53,7 @@ We also support limited customisation of the error codes and messages returned b
 | 400  | Request apiID does not match that in Definition! For Update operations these must match | Attempted a PUT operation using different api_ID's. Make sure the api_ID's are the same.                                                                                |
 | 400  | Request field is missing                                                                | Check if the request field is missing. If so, fill in the request field.                                                                                                |
 | 400  | Request ID does not match that in policy! For Update operations these must match        | Attempted a PUT operation using different policy ID's. Make sure both policy ID's are the same.                                                                         |
+| 400  | Request is too large                                                                   | The request body exceeds the configured size limit for the API endpoint.                                                                                                                           |
```
dcs3spp, line 56: @andyo-tyk Should the request is too large be HTTP status code 413?

```suggestion
| 413  | Request is too large                                                                   | The request body exceeds the configured size limit for the API endpoint.                                                                                                                           |
```

dcs3spp (changes requested):

tyk-docs/content/error-response-codes.md
andyo-tyk: It should be... but it isn't. :(
So here I'm documenting what the code actually returns, rather than what it _should_ return.
Ref: https://github.com/TykTechnologies/tyk/blob/29d50289c48ba2fccacba5306dc8fb408be15e85/gateway/mw_request_size_limit.go#L35
It's "on my list" to correct these inconsistencies... but as they would be breaking changes we have to handle them carefully.

dcs3spp: ok thanks @andyo-tyk

dcs3spp (approved):
lgtm

PR #2684, opened 2023-06-15
Tyk-Docs Release 5.1

dcs3spp:
PR for release-5.1 branch. It contains commits for:

- Updated redirect files for release 5.1, it contains the committed _redirects file.


dcs3spp:
@buger If this is ok can the release-5 branch be created now with this _redirects file? 

Making it stable is done on master branch by editing stable-updater.yaml GHAction. So this should be the last step that will make 5.1 live on netlify right?
https://github.com/TykTechnologies/tyk-docs/blob/master/.github/workflows/stable-updater.yaml#L12

buger (approved):

PR #2695, opened 2023-06-15
revert release versions dropdown so that it does not include release 5.1

dcs3spp:
remove 5.1 from release versions dropdown and _redirects


letzya:
shouldn't we remove the redirect as well? per https://github.com/TykTechnologies/tyk-docs/pull/2694/files

PR #2741, opened 2023-06-20
[DX-398] Update contribution guide

dcs3spp:
Update the Tyk Contribution Guide as follows:

- Explain that Tyklings can contribute by creating a PR branch
- Update usage example of `hugo new` with --configDir option from root folder of the repository.


dcs3spp:
@letzya

PR #2769, opened 2023-06-23
Revert "update release versions dropdown to include 5.1"

dcs3spp:
Reverts TykTechnologies/tyk-docs#2768


dcs3spp:
Fixed by applying leo's / fix

PR #2801, opened 2023-06-30
[TT-5866] Correcting typos, minor improvements to wording

andyo-tyk:
This probably needs updating (given it references Tyk 2.7) but taking this opportunity for a simple refresh.

No technical changes, does not need peer review, can be published immediately.

Prompted by TT-5866.

### DIFF
--- a/tyk-docs/content/planning-for-production.md
+++ b/tyk-docs/content/planning-for-production.md
@@ -13,33 +13,33 @@ aliases:
 
 So you want to deploy Tyk to production?
 
-There's a few things worth noting that can ensure your the performance of your Tyk Gateway nodes. Here's some of the basic things we do for load testing to make sure machines don't run out of resources.
+There's a few things worth noting that can ensure high performance of your Tyk Gateway nodes. Here are some of the basic things we do for load testing to make sure machines don't run out of resources.
 
 ### Performance Expectations
 
-Our performance testing plan focused on replicating setup of our customers, and try not to optimise for "benchmarks": so no supercomputers and no sub-millisecond inner DC latency. Instead, we were testing on super-low performance 2 CPU Linode machine, with 50ms latency between Tyk and upstream. For testing, we used Tyk Gateway in Multi-Cloud mode, with default config. Test runner was using [Locust][2] framework and [Boomer][3] for load generation.
+Our performance testing plan focused on replicating the setup of our customers, and tried not to optimise for "benchmarks": so no supercomputers and no sub-millisecond inner DC latency. Instead, we tested on a super-low performance 2 CPU Linode machine, with 50ms latency between Tyk and upstream. For testing, we used Tyk Gateway in Multi-Cloud mode, with default config. Test runner was using [Locust][2] framework and [Boomer][3] for load generation.
 
-With the optimisations outlined below, and using our distributed rate limiter, we can easily handle ~3,000 requests per second with analytics, key authentication, and quota checks enabled.
+With the optimisations outlined below, and using our distributed rate limiter, Tyk can easily handle ~3,000 requests per second with analytics, key authentication, and quota checks enabled.
 
-In the test below, Tyk is evaluating each request through its access control list, rate limiter, quota evaluator, and analytics recorder across a single test token and still retains a latency firmly under 70 milliseconds:
+In the test below, Tyk v2.7 evaluates each request through its access control list, rate limiter, quota evaluator, and analytics recorder across a single test token and still retains a latency firmly under 70 milliseconds:
 
 {{< img src="/img/diagrams/deployGraph.png" alt="Tyk 2.7 performance" >}}
 
 #### Performance changes based on use case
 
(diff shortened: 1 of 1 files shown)


tyk-docs/content/planning-for-production.md
```diff
@@ -13,33 +13,33 @@ aliases:
...
 
 {{< img src="/img/diagrams/deployGraphNoRateLimitQuota.png" alt="Tyk 2.7 performance" >}}
 
-Here we have pushed the test to 3,000 requests per second, and we can see that Tyk copes just fine - a with a few spikes past the 100ms line, we can clearly see solid performance right up to 3,000 requests per second with acceptable latency.
+Here we have pushed the test to 3,000 requests per second, and we can see that Tyk copes just fine. With only a few spikes past the 100ms line, we can clearly see solid performance right up to 3,000 requests per second with acceptable latency.
 
 #### Vanilla Tyk
 
-Now if you were to just test Tyk as a pass-through auth proxy, we can see that 4k requests per second is easily handled:
+If you were to just test Tyk as a pass-through auth proxy, we can see that 4k requests per second (rps) is easily handled:
```
letzya:
```suggestion
If you were to just test Tyk as a pass-through auth proxy, we can see that 4k RPS (requests per second) is easily handled:
```

tyk-docs/content/planning-for-production.md
```diff
@@ -13,33 +13,33 @@ aliases:
...
 
 {{< img src="/img/diagrams/deployGraphVanilla.png" alt="Tyk 2.7 performance" >}}
 
-This configuration has analytics recording disabled, but we are still authenticating the inbound request. As we can see we easily handle the 4k request per second mark, and we can go further with some more optimisations.
+This configuration has analytics recording disabled, but we are still authenticating the inbound request. As you can see, we easily handle the 4k rps mark and we can go further with more optimisation.
```
letzya:
```suggestion
This configuration has analytics recording disabled, but we are still authenticating the inbound request. As you can see, we easily handle the 4k RPS mark and we can go further with more optimisation.
```

tyk-docs/content/planning-for-production.md
```diff
@@ -161,40 +159,38 @@ Override your `systemd` unit files for each of the Tyk services using `systemctl
...
-```{.copyWrapper}
+```
 systemctl restart tyk-gateway.service
 ```
 
 #### Docker
 
-You may set ulimits in a container using the `--ulimit` option. See
-<a href="https://docs.docker.com/engine/reference/commandline/run/#set-ulimits-in-container---ulimit" target="_blank">Docker documentation</a> for setting ulimits.
+You may set ulimits in a container using the `--ulimit` option. See [Docker documentation]({{< ref "https://docs.docker.com/engine/reference/commandline/run/#set-ulimits-in-container---ulimit" >}}) for detail on setting ulimits.
```
letzya:
```suggestion
You may set *ulimits* in a container using the `--ulimit` option. See *Docker* documentation for detail on [setting *ulimits*]( https://docs.docker.com/engine/reference/commandline/run/#set-ulimits-in-container---ulimit)
```

letzya (dismissed):
please commit my suggestions

tyk-docs/content/planning-for-production.md
andyo-tyk: Typically, when introducing terminology, you would write it out in full and then indicate the abbreviation. Agree the abbreviation should be capitalised.
```suggestion
If you were to just test Tyk as a pass-through auth proxy, we can see that 4k requests per second (RPS) is easily handled:
```

letzya: this is trivial information in this context, it's a lot more known and clear than `pass-through auth proxy`, it's almost odd to even add `request per second`

andyo-tyk:
@letzya I've committed two, but the other I disagree with and have proposed a slight tweak. If you insist we can go with yours. Please advise!

tyk-docs/content/planning-for-production.md
andyo-tyk: That's incidental (though I agree) - but if we're going to use an abbreviation it's good practice to expand it to avoid ambiguity. I'm not going to die on that hill, however, so will commit your suggestion. This page needs an overhaul/update anyway.

tyk-docs/content/planning-for-production.md
```diff
@@ -161,40 +159,38 @@ Override your `systemd` unit files for each of the Tyk services using `systemctl
...
+If you are not using systemd or Docker, please consult your Operating System documentation for controlling the number of File Descriptors available for your process.
 
 ### File modification at runtime
 
 Understanding what files are created or modified by the Dashboard and Gateway during runtime can be important if you are running infrastructure orchestration systems such as Puppet, which may erroneously see such changes as problems which need correcting.
 
 *   Both the Gateway and Dashboard will create a default configuration file if one is not found.
 *   Dashboard will write the licence into the configuration file if you add it via the UI.
-*   From v2.3 onwards it is possible for a Dashboard to remotely change the config of a Gateway, which will cause the Gateway's configuration file to update.
+*   From Tyk v2.3 onwards it has been possible for a Dashboard to remotely change the config of a Gateway: this will cause the Gateway's configuration file to update.
```
dcs3spp:
```suggestion
*   From Tyk v2.3 onwards it has been possible for a Dashboard to remotely change the config of a Gateway. This will cause the Gateway's configuration file to update.
```

dcs3spp (approved):
lgtm

PR #2825, opened 2023-07-04
fix weighting so that mcdb 2.3 appears first in list of release notes

dcs3spp:
adjust weighting for mcdb release note 2.3 so that appears first in list of mcdb releases


dcs3spp:
netlify checks failed

PR #2989, opened 2023-07-27
[TT-8959/TT-9613] Add endpoint level cache timeout doc

furkansenharputlu:
This PR adds documentation for the endpoint level cache timeout value.


dcs3spp:
@furkansenharputlu PR looks good to me. Please label with targeted release.

jeffy-mathew (approved):

andyo-tyk:
I think this has been added to the wrong place in the page - it's been added to "Selective caching by body value", it should be in "Selective caching by endpoint" as that is the use case where a user might be particularly interested in setting different timeouts for different endpoints.

dcs3spp:
@furkansenharputlu per comment made by @andyo-tyk should the timeout field moved into the cache by endpoint section?

dcs3spp (changes requested):
@furkansenharputlu per comment made by @andyo-tyk should the timeout field moved into the cache by endpoint section?

furkansenharputlu:
Imho, I think `timeout` belongs to the correct place. Things should move as a block, the `timeout` is the inner piece and it belongs to the correct configuration example.

andyo-tyk:
I will create a new Jira ticket for a restructuring of this page - in fact there should be a page dedicated to per-endpoint caching so will probably move all of that to a new page.

Please feel free to approve this PR, @dcs3spp

dcs3spp (approved):

PR #3046, opened 2023-08-02
[DX-581] Update secret docs for passing target url from env

joshblakeley:
It was undocumented that you can pass target url or listen path with the env var TYK_SECRET_{some string} and have that picked up in API definitions with env://{some string}

This is important functionality for users with shared api definitions that land on multiple clusters of gateways for example

### DIFF
--- a/tyk-docs/content/tyk-configuration-reference/kv-store.md
+++ b/tyk-docs/content/tyk-configuration-reference/kv-store.md
@@ -99,6 +99,15 @@ TYK_GW_USEDBAPPCONFIGS
 TYK_GW_POLICIES_POLICYSOURCE
 ```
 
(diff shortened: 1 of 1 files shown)


tyk-docs/content/tyk-configuration-reference/kv-store.md
```diff
@@ -99,6 +99,15 @@ TYK_GW_USEDBAPPCONFIGS
 TYK_GW_POLICIES_POLICYSOURCE
 ```
 
+Additionally if you with to override either the listen path or target URL in an API definition you can store a number of secrets in this format:
```
dcs3spp:
```suggestion
Additionally if you wish to override either the listen path or target URL in an API definition you can store a number of secrets in this format:
```

dcs3spp (approved):
lgtm

PR #3092, opened 2023-08-08
[DX-606]adding colors variables that will be used for rebranding

yurisasuke:
DX-606
---


dcs3spp (approved):
lgtm

PR #3115, opened 2023-08-10
Fix LTS target for docs update

titpetric:
For the LTS target the value of `5.0 LTS` was used in the branch name, causing a failure. This PR fixes it, tested via #3114 on exp/ location. Doesn't require documentation changes, ready for merge.


dcs3spp (approved):
Thanks @titpetric, good catch. Yep docs release branch is release-5 and source branch uses release-5-lts

PR #3129, opened 2023-08-14
[DX-630] Improved clarity for key lifetime

andyo-tyk:
Improved visibility and clarity of the default value for session_lifetime and that this means an infinite lifetime (keys will not be deleted)


dcs3spp (changes requested):
PR lgtm Andy. Once peer reviewed we can merge and release

dcs3spp (approved):

lghiur (approved):

PR #3211, opened 2023-08-31
[DX-649]Home tab should be selected when at home

yurisasuke:

DX-649
---
### Preview Link
https://deploy-preview-3211--tyk-docs.netlify.app/docs/nightly/

### DIFF
--- a/tyk-docs/themes/tykio/layouts/partials/tabs_menu.html
+++ b/tyk-docs/themes/tykio/layouts/partials/tabs_menu.html
@@ -6,12 +6,12 @@
 
       {{- range .Site.Data.menu.menu -}}
         {{- $selected := false -}}
-        {{- $not_home_path := false -}}
+        {{- $path_not_empty := false -}}
         {{- $path := replace .path "/" "" -}}
         {{- if eq $path $pagePath -}}
         {{- $selected = true -}}
           {{- if ne $path "" -}}
-            {{- $not_home_path = true -}}
+            {{- $path_not_empty = true -}}
           {{- end -}}
         {{- end -}}
         {{- if (isset . "menu") -}}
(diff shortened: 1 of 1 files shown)


tyk-docs/themes/tykio/layouts/partials/tabs_menu.html
```diff
@@ -6,75 +6,76 @@
...
         {{- $selected := false -}}
-        {{- $not_home_path := false -}}
+        {{- $is_home := eq (lower .title) "home" -}}
+        {{- $path_not_empty := false -}}
         {{- $path := replace .path "/" "" -}}
         {{- if eq $path $pagePath -}}
         {{- $selected = true -}}
-          {{- if ne $path "" -}}
-            {{- $not_home_path = true -}}
+          {{- if or (ne $path "") $is_home -}}
```
dcs3spp, line 14: So the or condition means _if the path is not empty OR we are on page "home"_, yes?

If the OR condition is true then $path_not_empty gets set to true and the home page link is activated

dcs3spp (commented):
So the OR condition means 

- if the path is not empty OR we are on page "home", yes?

If the OR condition is true then _$path_not_empty_ gets set to true and the home page link gets activated

tyk-docs/themes/tykio/layouts/partials/tabs_menu.html
yurisasuke: @dcs3spp  $path_not_empty is to check that the menu item we are currently looping does actually have a path. You remember we have menuItems that have no paths. The ones that we labelled as show: false.
$is_home - Since home uses an empty path we want to exclude it from being checked as empty.

dcs3spp (approved):
Approved! PR looks good to me Keith.

PR #3221, opened 2023-09-04
remove SVG error in docs's hugo build

yurisasuke:

DX-198
---
### Description
- Since gif and svg don't need to be resized we should exclude them from the hugo build error. that shows the message such as `Wrong image type /docs/nightly/img/2.10/tyk_operator2.svg`


dcs3spp (changes requested):
Hi @Keithwachira How does this fix the issue?

yurisasuke:
> Hi @Keithwachira How does this fix the issue?



> Hi @Keithwachira How does this fix the issue?

I added this line  `{{ else if in (slice "svg" "gif" ) .MediaType.SubType }}` that checks if the image is a gif or svg and skips displaying a warning for such images

dcs3spp (approved):

PR #3280, opened 2023-09-22
[DX-698] fix issue with search where 3.1 is getting indexed

yurisasuke:

DX-698


dcs3spp (approved):
lgtm

PR #3311, opened 2023-10-03
fix: use official dynatrace sample token

patrickpichler:
The example token for dynatrace has been replaced with the official token as listed by the dynatrace docs.

https://www.dynatrace.com/support/help/dynatrace-api/basics/dynatrace-api-authentication


tbuchaillot (approved):
lgtm!

PR #3327, opened 2023-10-04
doc: Specyfing method for ignore/whitelist/blacklist

jakub-bochenski:
This works for me but is not documented.

Is this intended behavior?

### For internal users - Please add a Jira DX PR ticket to the subject!
<br>

---

### Preview Link

<br>

### Description

<br>

#### Screenshots (if appropriate)
<br>

## Checklist

- [ ] I have added a **preview link** to the PR description.
- [x] I have [reviewed the guidelines](https://github.com/TykTechnologies/tyk-docs/blob/master/CONTRIBUTING.md) for contributing to this repository.
- [x] I have [read the technical guidelines](https://github.com/TykTechnologies/tyk-docs/blob/master/CONTRIBUTING-TECHNICAL-GUIDE.md) for contributing to this repository.
- [x] Make sure you have started *your change* off *our latest `master`*.
- [ ] I **labelled** the PR

### DIFF
--- a/tyk-docs/content/shared/api-def-version.md
+++ b/tyk-docs/content/shared/api-def-version.md
@@ -105,6 +105,7 @@ Each entry in the ignored, blacklist and whitelist have the same specification. 
 ```{.copyWrapper}
 {
   "path": "SOME_PATH",
+  "method": "SOME_METHOD",
   "method_actions": {
     "METHOD": {
       "action": "ACTION_CODE",
@@ -119,6 +120,8 @@ Each entry in the ignored, blacklist and whitelist have the same specification. 
 ```
     
 You can set up the path to handle, and the action type. By default this should be set to `no_action` which means that Tyk will treat the path and method as-is without interference (i.e. if `no_action` is set, then `code`, `data`, `headers` and `body` are *not* used).
+
+If method is not specified it will apply to all methods.
     
 If you set `action` to `reply` Tyk will override the path and reply with settings defined in the remaining fields. This configuration can be used across all of the different path lists, and can be very useful when used in conjunction with the `white_list` access control list as if it is used exclusively it can be used to mock out an entire API.
     


jakub-bochenski:
I have no idea what the unchecked point on the checklist are about

jakub-bochenski:
Seems this is available via dashboard https://tyk.io/docs/img/2.10/ignore.png

letzya:
@jakub-bochenski Thank you for taking the time to create a pull request for our docs. We really appreciate your contribution!
We will test and confirm as I roughly remember you are correct. You'll see feedback from us in the next few days. 
Thanks again
Yaara

andyo-tyk:
Hi @jakub-bochenski,

Thank you for raising this, you've actually found a bug - this is not intended behaviour. The `method` field that you've used should be a private field not exposed in the API Definition - we'll be correcting this in an upcoming release. I'm sorry for any confusion or inconvenience caused.

Incidentally, the screenshot that you've included for Ignore actually refers to the case (capitalisation) of the endpoint.

Thank you for supporting Tyk!

jakub-bochenski:
@andyo-tyk  but this is actually useful for me.

How can I whitelist e.g. only a "GET" request then and not all methods?

andyo-tyk:
Hi @jakub-bochenski,

The whitelist (allow list) capability is described in the docs [here](https://tyk.io/docs/tyk-apis/tyk-gateway-api/api-definition-objects/versioning-endpoint/).

If you want to allow only calls to GET for an endpoint, then you need to configure GET in `method_actions`, for example:

```
"white_list": [
  {
    "path": "<YOUR_ENDPOINT>",
    "method_actions": {
      "GET": {
        "action": "no_action",
        "code": 200,
        "data": "",
        "headers": {}
      }
    }
  }
], 
```

Note that by setting `action` to `no_action`,  Tyk will treat the path and method as-is without interference (i.e. `code`, `data` and `headers` are not used).

I hope this helps you achieve what you're after.

jakub-bochenski:
@andyo-tyk  according to the linked documentation it will not work

> You can set up the path to handle, and the action type. By default this should be set to no_action which means that Tyk will treat the path and method as-is without interference (i.e. if no_action is set, then code, data, headers and body are not used).

So I would have to explicitly deny all possible methods other than what I want to allow

jakub-bochenski:
OK it seems to actually work in practice.

But the the documentation is unclear then.

jakub-bochenski:
This wording is very confusing 

> By default this should be set to no_action 

What does it actually mean? I don't think "no_action" is the default

andyo-tyk:
Hi @jakub-bochenski,

Thank you for the feedback on the documentation - I agree, it could be clearer.

Here we are setting up an Allow List - so anything not strictly listed will default to Deny.

With Tyk Classic APIs, the Mock Response is configured within the Allow List (can also be done within Deny and Ignore) - and the `action` field is used to configure how the mock should respond. By setting this to `no_action` we are instructing Tyk not to provide a mock response, but to allow the request to pass to the upstream.

I hope this now makes sense and that you're able to configure the Allow/White listing how you need it.

Thanks again for supporting Tyk!

PR #3392, opened 2023-10-17
rename release notes template file to include suffix

dcs3spp:
rename release notes template file to include suffix


dcs3spp:
Rename release notes template file so it include .md suffix

PR #3442, opened 2023-10-26
Footer redesign for homepage

yurisasuke:
[Preview link](https://deploy-preview-3442--tyk-docs.netlify.app/docs/nightly)


tyk-docs/assets/scss/_bespoke.scss
```diff
@@ -633,7 +633,7 @@ input {
 input,
 textarea {
   color: #5a5a68;
-  padding-left: 18px;
+  padding-left: 7px!important;
```
ifrim: `!important` should be generally avoided because it breaks the cascading nature of css. It should be used only when there is no other possibility. For example when trying to override some 3rd-party library css and there's no other way.

It is especially problematic when using it with general selectors like in this case with `textarea`. Now if you need a textarea where you don't want to have padding left of 7px it is hard to overwrite that css. Also this would override all existing textareas, wherever they are used, which might have unintended consequences.

ifrim (approved):
`!important` is used in way too many places which breaks the cascading nature of css and will make things harder to override and maintain in the future. As a best practice it should be avoided. See the best practices section from this page for a more detailed explanation https://developer.mozilla.org/en-US/docs/Web/CSS/important#best_practices

Otherwise looks good.

tyk-docs/assets/scss/_bespoke.scss
yurisasuke: > `!important` should be generally avoided because it breaks the cascading nature of css. It should be used only when there is no other possibility. For example when trying to override some 3rd-party library css and there's no other way.
> 
> It is especially problematic when using it with general selectors like in this case with `textarea`. Now if you need a textarea where you don't want to have padding left of 7px it is hard to overwrite that css. Also this would override all existing textareas, wherever they are used, which might have unintended consequences.

Thank you @ifrim .This information is really helpful. I have now removed important in all places that it was not needed.

yurisasuke:
> `!important` is used in way too many places which breaks the cascading nature of css and will make things harder to override and maintain in the future. As a best practice it should be avoided. See the best practices section from this page for a more detailed explanation https://developer.mozilla.org/en-US/docs/Web/CSS/important#best_practices
> 
> Otherwise looks good.

Thank you so much for the review. And for the important information.

PR #3517, opened 2023-11-08
add fixxing for list element to add padding

yurisasuke:
Add padding to li element as reported by andy

>. I have removed important to make sure it does not override the already exiting css . This make sure that Li in content have a padding of 15px

[Preview link](https://deploy-preview-3517--tyk-docs.netlify.app/docs/nightly/getting-started/key-concepts/high-level-concepts/)


dcs3spp (approved):
PR lgtm

PR #3551, opened 2023-11-14
remove explore from assets as it is a duplicate and should be in static

yurisasuke:
- Images used in css and html should be in static and not assets.
[Preview Link](https://deploy-preview-3551--tyk-docs.netlify.app/docs/nightly/)


dcs3spp (approved):
PR LGTM

PR #3650, opened 2023-11-28
[TT-10597] Update documentation for 5.2

buger:
Triggered by: titpetric

Included:

Tyk Gateway: true
Tyk Dashboard: true
Tyk MDCB false
Tyk Pump false

Intended for: 5.2

Note: Release 5.2.4

JIRA: https://tyktech.atlassian.net/browse/TT-10597


titpetric:
We won't be publishing the oauth tokens endpoint at this point, subject to further refinement

PR #3728, opened 2023-12-11
[DX-886] remove hotjar as we no longer use it

yurisasuke:
[DX-886]
Remove hotjar as we no longer use it. We use full story instead. This is as an effort to remove all javascript that we do not use as they slow down our website

[DX-886]: https://tyktech.atlassian.net/browse/DX-886?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


letzya (approved):

letzya:
@buger FYI

PR #3766, opened 2023-12-18
[DX-917]generate url mapping  for each version in json format

yurisasuke:
[DX-917]
**Context:**
We have an issue in which, if we move a page to another file path, and add an alias, then if the user wants to see that page in a previous version, they will get 404 since that URL does not exist in that version, only the URL that is now an alias. 

**Purpose of this pull request**: - To fix this issue we generate all pages of the pages of each version and map the url in the old struture to the new structure. This means we will be able to move bwtween versions even thought they use different file structure.

**Purpose of the  List.page.url.json** : This is the file we use to generate all the urls of a particular versions and all its aliases . Each alias will also have a url to its parent.

[DX-917]: https://tyktech.atlassian.net/browse/DX-917?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dcs3spp (approved):
PR lgtm

letzya:
Should I be able to see previous versions for this url https://deploy-preview-3766--tyk-docs.netlify.app/docs/nightly/debugging-series/debugging-series/?

letzya:
@Keithwachira could you please add a few URLs to show how this was tested?
I tried https://deploy-preview-3766--tyk-docs.netlify.app/docs/nightly/release-notes/version-4.0/ but when I change it to a previous version i get 404, so not sure how to test it

letzya:
Error in the CI
> 5:21:37 PM:   The Node.js version is 12.18.0 but the plugin "@netlify/plugin-lighthouse" requires >=14.15 <20

image

letzya (approved):

PR #3801, opened 2023-12-19
Implement version selector master dynamic and also solve backward file structure

yurisasuke:
We have now disabled version in which a given page does not appear in. This is to prevent users from clicking pages that will return a 404 anyway. We also want to make sure that if a previous page used a different url users will be able to access that page. For example let say a page was introduced in version 5.1 of the docs, we have disabled any version below version 5.1 
[DX-908]
**FYI Currently only version 4.2 and master data have been added**

[DX-908]: https://tyktech.atlassian.net/browse/DX-908?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

https://deploy-preview-3801--tyk-docs.netlify.app/docs/nightly/product-stack/tyk-gateway/advanced-configurations/distributed-tracing/open-tracing/jaeger/


yurisasuke:
replaced by https://github.com/TykTechnologies/tyk-docs/pull/3989

PR #3848, opened 2024-01-04
[TT-10749, DX-949] Update rate limit docs with leaky bucket option

titpetric:
### Preview Link

<br>

### Description

This PR adds documentation for the new leaky bucket rate limiting option in Gateway 5.3 release.

It updates some false documentation:

- DRL implements token bucket and not leaky bucket (errata),
- RRL implements sliding window log (as opposed to sliding window counter), be accurate in naming

[Preview link](https://deploy-preview-3848--tyk-docs.netlify.app/docs/nightly/getting-started/key-concepts/rate-limiting/)

<br>

#### Screenshots (if appropriate)
<br>

## Checklist

- [ ] I have added a **preview link** to the PR description.
- [ ] I have [reviewed the guidelines](https://github.com/TykTechnologies/tyk-docs/blob/master/CONTRIBUTING.md) for contributing to this repository.
- [ ] I have [read the technical guidelines](https://github.com/TykTechnologies/tyk-docs/blob/master/CONTRIBUTING-TECHNICAL-GUIDE.md) for contributing to this repository.
- [ ] Make sure you have started *your change* off *our latest `master`*.
- [ ] I **labelled** the PR

PR #3917, opened 2024-01-16
Merging to release-5: [TT-10749, DX-949] Update rate limit docs with leaky bucket option (#3848)

buger:
[TT-10749, DX-949] Update rate limit docs with leaky bucket option (#3848)

* Update rate limit docs with leaky bucket option


dcs3spp:
Closing since feature not on 5 LTS

PR #3962, opened 2024-01-19
[TT-7739] Update documentation for master

buger:
Triggered by: titpetric

Included:

Tyk Gateway: true
Tyk Dashboard: true
Tyk MDCB true
Tyk Pump true

Intended for: master
Config info generator branch: keep-sorting-fix-env

Note: Verifying change for TT-7739

JIRA: https://tyktech.atlassian.net/browse/TT-7739

### DIFF
--- a/tyk-docs/content/shared/gateway-config.md
+++ b/tyk-docs/content/shared/gateway-config.md
@@ -1,35 +1,35 @@
 ### hostname
-EV: <b>TYK_GW_HOSTNAME</b><br />
+ENV: <b>TYK_GW_HOSTNAME</b><br />
 Type: `string`<br />
 
 Force your Gateway to work only on a specific domain name. Can be overridden by API custom domain.
 
 ### listen_address
-EV: <b>TYK_GW_LISTENADDRESS</b><br />
+ENV: <b>TYK_GW_LISTENADDRESS</b><br />
 Type: `string`<br />
 
 If your machine has multiple network devices or IPs you can force the Gateway to use the IP address you want.
 
 ### listen_port
-EV: <b>TYK_GW_LISTENPORT</b><br />
+ENV: <b>TYK_GW_LISTENPORT</b><br />
 Type: `int`<br />
 
 Setting this value will change the port that Tyk listens on. Default: 8080.
 
 ### control_api_hostname
-EV: <b>TYK_GW_CONTROLAPIHOSTNAME</b><br />
+ENV: <b>TYK_GW_CONTROLAPIHOSTNAME</b><br />
 Type: `string`<br />
 
 Custom hostname for the Control API
 
 ### control_api_port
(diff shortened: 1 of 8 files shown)


titpetric:
The change of `EV` to `ENV` verified with this PR and will be in effect for the official docs release.

Best,
Tit

letzya:
FYI @dcs3spp https://github.com/TykTechnologies/tyk-docs/pull/3962#issuecomment-1901057635

tyk-docs/content/shared/gateway-config.md
```diff
@@ -350,36 +351,38 @@ If you set this value to `true`, then the id parameter in a stored policy (or im
 This option should only be used when moving an installation to a new database.
 
 ### policies.policy_path
-EV: <b>TYK_GW_POLICIES_POLICYPATH</b><br />
+ENV: <b>TYK_GW_POLICIES_POLICYPATH</b><br />
 Type: `string`<br />
 
-This option is used for storing a policies  if `policies.policy_source` is set to `file`.
-it should be some existing file path on hard drive
+This option only applies in OSS deployment when the `policies.policy_source` is either set
```
letzya, line 357: @tit why do we only explain the oss. One might wonder what is the "other" or "unnamed" option which we don't talk about. What is the value of that be? @dcs3spp could you please advise with the wording?

tyk-docs/content/shared/gateway-config.md
```diff
@@ -636,14 +639,26 @@ This is to ensure visibility for the management node across all APIs.
 ### auth_override
 This is used as part of the RPC / Hybrid back-end configuration in a Tyk Enterprise installation and isn’t used anywhere else.
 
+### enable_leaky_bucket_rate_limiter
+ENV: <b>TYK_GW_ENABLELEAKYBUCKETRATELIMITER</b><br />
+Type: `bool`<br />
+
+EnableLeakyBucketRateLimiter enables leaky bucket rate limiting.
```
letzya, line 646: @titpetric @lghiur  adding `EnableLeakyBucketRateLimiter` makes no sense for the user. Either use the env var notation or the json

tyk-docs/content/shared/gateway-config.md
```diff
@@ -636,14 +639,26 @@ This is to ensure visibility for the management node across all APIs.
...
+
+LeakyBucket will delay requests so they are processed in a FIFO
+style queue, ensuring a constant request rate and smoothing out
+traffic spikes. This comes at some cost to gateway instances, as
+the connections would be held for a longer time, instead of
```
letzya, line 651: @lghiur a bit confusing 
> This comes at some cost to gateway instances, as the connections would be held for a longer time, instead of  blocking the requests when they go over the defined rate limits.

it has pros and cons in the sentence also it says it "comes at some cost"

PR #3986, opened 2024-01-22
[DX-908]Generate url availability file

yurisasuke:
This pr contains the github action that will be run to produce the json file that will be used to create the dynamic version selector.

### DIFF
--- a/scripts/page_available_since.py
+++ b/scripts/page_available_since.py
@@ -54,10 +54,11 @@ versions = [
         "branch": "release-3-lts"
     },
 
-    {"path": "/docs/nightly/",
-     "name": "Nightly",
-     "branch": "master"
-     }
+    {
+        "path": "/docs/nightly/",
+        "name": "Nightly",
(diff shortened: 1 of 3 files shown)


scripts/page_available_since.py
```diff
@@ -0,0 +1,162 @@
...
+        "name": "3.1",
+        "branch": "release-3.1"
+    },
+    {
+        "path": "/docs/3-lts/",
+        "name": "3 LTS",
+        "branch": "release-3-lts"
+    },
+
+    {"path": "/docs/nightly/",
```
dcs3spp: very minor comment, for consistency start "path" key on new line

dcs3spp (approved):
PR LGTM and approving. As discussed in meet add comments for block of codes that stores similar and diff paths.

scripts/page_available_since.py
yurisasuke: done

PR #3640, opened 2023-11-27
[Dx 720] - Update Release Landing Page With Latest Component Versions

dcs3spp:
[DX-720](https://tyktech.atlassian.net/browse/DX-720)

[preview - release landing page](https://deploy-preview-3640--tyk-docs.netlify.app/docs/nightly/developer-support/tyk-release-summary/overview)

Update Release landing page to include latest version info, with links to GHub and DockerHub images

[DX-720]: https://tyktech.atlassian.net/browse/DX-720?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #4029, opened 2024-01-25
Merging to release-5.1: [Dx 720] - Update Release Landing Page With Latest Component Versions (#3640)

buger:
[Dx 720] - Update Release Landing Page With Latest Component Versions (#3640)

Update release landing page with latest component versions

### DIFF
--- a/tyk-docs/assets/scss/main.scss
+++ b/tyk-docs/assets/scss/main.scss
@@ -1,9 +1,6 @@
 @import 'variables';
-<<<<<<< HEAD
-=======
 @import 'home';
 @import 'releases';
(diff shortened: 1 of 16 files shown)


letzya:
I think it'll be too much to handle since we'll need to update this version and the latest for every release. What do you think? will 5.2 be enough for now?

dcs3spp:
> I think it'll be too much to handle since we'll need to update this version and the latest for every release. What do you think? will 5.2 be enough for now?

Ok, thanks @letzya I'll include it in 5.2 for now. I'll close this PR and add a link to the PR in the related JIRA issue so that can reopen if necessary

PR #5520, opened 2024-10-07
Update tyk-oss-chart.md

gecube:
fix the docs helm

There was missing two lines pointing to a particular key in `values.yaml`


buraksekili (approved):
thanks @gecube for fixing this!

sharadregoti (approved):

PR #5600, opened 2024-10-15
[DX-1704]add prettier to the code  for formatting

yurisasuke:
To make sure we have consistent formatting of javascript,markdown and html we have introduced prettier to help us with the formatting.

Prettier completes with errors if you have invalid html or json.

In this Pr i have also fixed html errors that existed in the docs e.g in badge.html

[DX-1704]

[DX-1704]: https://tyktech.atlassian.net/browse/DX-1704?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.prettierrc
```diff
@@ -0,0 +1,32 @@
+{
+  "plugins": [
+    "prettier-plugin-go-template"
+  ],
+  "goTemplateBracketSpacing": true,
+  "printWidth": 120,
+  "proseWrap": "always",
```
letzya, line 6: @Keithwachira is it ok we wrap for all file types? shouldn't it be for md only?

verions.json
```diff
@@ -2,7 +2,7 @@
   {
```
titpetric: should this file be named `versions.json`?

.prettierrc
```diff
@@ -0,0 +1,32 @@
...
+  "tabWidth": 2,
+  "overrides": [
+    {
+      "files": [
+        "*.html"
+      ],
+      "options": {
+        "parser": "go-template",
+        "goTemplateBracketSpacing": true,
```
letzya, line 16: do we need this if it's in the global scope?

.prettierrc
```diff
@@ -0,0 +1,32 @@
...
+      }
+    },
+    {
+      "files": [
+        "*.js",
+        "*.ts"
+      ],
+      "options": {
+        "useTabs": true,
+        "printWidth": 120,
```
letzya: ame - in global scope so it's redundant

.prettierrc
```diff
@@ -0,0 +1,32 @@
...
+        "singleQuote": false
+      }
+    }
+  ]
+}
```
titpetric: Mising EOL as github reports

titpetric (approved):
LGTM, the only NIT I have is this could likely be extended for `*.(json,yml,yaml)`

.github/workflows/prettier.yml
```diff
@@ -0,0 +1,50 @@
...
+      - name: Get changed files
+        id: changed-files
+        uses: tj-actions/changed-files@v45
+
+      - name: Lint all changed files
+        env:
+          ALL_CHANGED_FILES: ${{ steps.changed-files.outputs.all_changed_files }}
+        run: |
+          for file in ${ALL_CHANGED_FILES}; do
+            npx prettier --write "$file"
```
letzya: do you mid if we add explicit call to the config file? 

`npx prettier --config .prettierrc --write "$file"`

letzya:
```suggestion
            npx prettier --config .prettierrc --write "$file"
```

.github/workflows/prettier.yml
```diff
@@ -0,0 +1,50 @@
...
+      - uses: stefanzweifel/git-auto-commit-action@v5
+        with:
+          commit_message: save formatted content
+
+      - name: Check all changed files are linted
+        env:
+          ALL_CHANGED_FILES: ${{ steps.changed-files.outputs.all_changed_files }}
+        run: |
+          for file in ${ALL_CHANGED_FILES}; do
+            npx prettier --check "$file"
```
letzya:
```suggestion
            npx prettier --config .prettierrc --check "$file"
```

.github/workflows/prettier.yml
```diff
@@ -0,0 +1,59 @@
+# This GitHub Action is triggered on pull requests to:
+# 1. Lint changed files using Prettier.
+# 2. Automatically install Prettier and its Go template plugin.
+# 3. Identify all files changed in the pull request.
+# 4. Format the changed files according to Prettier's rules, respecting the .prettierignore file.
+# 5. Commit any changes made to the files back to the repository.
+# 6. Optionally, check that the files are properly formatted after linting.
+#
+# This action helps ensure code quality and consistency in documentation and source files.
+name: "Lint the docs with prettier"
```
letzya:
```suggestion
name: "PR files Formatting"
```

.github/workflows/prettier.yml
```diff
@@ -0,0 +1,59 @@
+# This GitHub Action is triggered on pull requests to:
+# 1. Lint changed files using Prettier.
```
letzya:
```suggestion
# 1. Format changed files using Prettier.
```

.github/workflows/prettier.yml
```diff
@@ -0,0 +1,59 @@
...
+#
+# This action helps ensure code quality and consistency in documentation and source files.
+name: "Lint the docs with prettier"
+on:
+  pull_request:
+
+jobs:
+  docs_linter:
+    runs-on: ubuntu-latest
+    name: Lint the docs with prettier
```
letzya:
```suggestion
    name: Format PR files with Prettier
```

.prettierrc
yurisasuke: > @Keithwachira is it ok we wrap for all file types? shouldn't it be for md only?

I think we should wrap all file types to make sure readability and formatting is applied in all our files types instead of md files alone.

.prettierrc
yurisasuke: > do we need this if it's in the global scope?

Since the parser is for go templates I have moved it to the html parts and removed it from global.

.prettierrc
yurisasuke: > ame - in global scope so it's redundant

@letzya  I have now removed the redundant code.

.github/workflows/prettier.yml
yurisasuke: > do you mid if we add explicit call to the config file?
> 
> `npx prettier --config .prettierrc --write "$file"`

I have now added the explicit call to config file

.prettierrc
yurisasuke: > Mising EOL as github reports

I have now added it

verions.json
yurisasuke: > should this file be named `versions.json`?

Let me create a new pr to rename the file. (As it the file already existed in the docs)

letzya (approved):

PR #6179, opened 2025-03-19
add images to docs so that we can control them

yurisasuke:
Currently the swagger images are hosted on postman. However we want to host them here in the docs so that we can control them. 

This pr ads them to the docs so that they can be referenced by the dashboard and gateway swagger


sharadregoti (changes requested):
Add images for MDCB and Developer Portal (I have asked @jen to create it)

yurisasuke:
> Add images for MDCB and Developer Portal (I have asked @jen to create it)

Did she give a timeline onto when she might finish working on them

sharadregoti:
By tomorrow, we should get the images.

sharadregoti (approved):

